PR #15, opened 2024-07-29
fix: make getInscriptions to return txid and vout object

jrwbabylonlab:
This PR address some of the comments from previous PR review: https://github.com/babylonchain/simple-staking/pull/376

### DIFF
--- a/src/utils/wallet/providers/okx_wallet.ts
+++ b/src/utils/wallet/providers/okx_wallet.ts
@@ -179,6 +179,11 @@ export class OKXWallet extends WalletProvider {
     if (!this.okxWalletInfo) {
       throw new Error("OKX Wallet not connected");
     }
+    if (this.networkEnv !== Network.MAINNET) {
+      throw new Error(
+        "Inscriptions are only available on OKX Wallet BTC mainnet",
+      );
+    }
     // max num of iterations to prevent infinite loop
     const MAX_ITERATIONS = 100;
     // Fetch inscriptions in batches of 100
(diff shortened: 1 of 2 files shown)


gbarkhatov (approved):

jeremy-babylonlabs (approved):

src/utils/wallet/providers/okx_wallet.ts
```diff
@@ -176,23 +175,46 @@ export class OKXWallet extends WalletProvider {
   };
 
   // Inscriptions are only available on OKX Wallet BTC mainnet (i.e okxWallet.bitcoin)
-  getInscriptions = async (): Promise<Inscription[]> => {
+  getInscriptions = async (): Promise<InscriptionIdentifier[]> => {
```
vitsalis, line 178: Given that we replicate this function across two different wallets, can we have a util that does that?

src/utils/wallet/providers/okx_wallet.ts
```diff
@@ -176,23 +175,46 @@ export class OKXWallet extends WalletProvider {
   };
 
   // Inscriptions are only available on OKX Wallet BTC mainnet (i.e okxWallet.bitcoin)
```
vitsalis, line 177: Should we add a check on whether we are on `mainnet` or not? If not, we should throw an error or something.

src/utils/wallet/providers/onekey_wallet.ts
```diff
@@ -165,20 +164,43 @@ export class OneKeyWallet extends WalletProvider {
   }
 
   // Inscriptions are only available on oneKey Wallet BTC mainnet
```
vitsalis, line 166: Same here

vitsalis (approved):

PR #19, opened 2024-07-29
feature: Geo blocking

### LINKED ISSUE #5: Geo-blocking
Add an ability to inform users about geo-blocking

gbarkhatov:
Besides the previous chain implementation split

```mermaid
graph TD
    subgraph "API Layer"
        A[healthCheckClient.ts] -->|fetchHealthCheck| B(HealthCheckResponse)
    end
    
    subgraph "Service Layer"
        C[healthCheckService.ts] -->|getHealthCheck| D(HealthCheckResult)
        D --> E{HealthCheckStatus}
        D --> F(GEO_BLOCK_MESSAGE)
        D --> G(API_ERROR_MESSAGE)
    end
    
    subgraph "Hook Layer"
        H[useHealthCheck.ts] -->|useQuery|getHealthCheck
        H -->|handleError| I[useError from ErrorContext]
        H --> J{isApiNormal}
        H --> K{isBlocked}
        H --> L{apiMessage}
    end
    
    subgraph "Components"
        N[ConnectSmall.tsx] -->|apiAvailable| H
        O[Staking.tsx] -->|apiAvailable| H
    end

    A --> C
    C --> H
    I --> H
```

So basically
1. API fetch is one thing `healthCheckClient.ts`, just calls axios to fetch the data, but has a unique function condition if it's `451`. It is not aware of `geoBlocking`, it is aware of `451` code and can return that
2.  Service is another `healthCheckService`. It is assigning whether the response is `normal`, `error`, `geoBlocked`. It can also return whether or not the result is geo blocked. Based on `HealthCheckStatus.GeoBlocked` not on `451`
3. Custom hook. `useHealthCheck` uses `react query` and our custom error context
4. Components use this hook to get the data

There are a lot of steps to go from there, but it's the general approach we will use (after the polishing, conventions, etc) to
1. Stop API stuff have the business logic inside
2. Separate business logic based on the API stuff
3. Properly work in different components

Video: https://youtu.be/u_eZls4mEpA
Closes #5


src/app/api/healthCheckClient.ts
```diff
@@ -0,0 +1,19 @@
...
+}
+
+export const fetchHealthCheck = async (): Promise<HealthCheckResponse> => {
+  const response = await axios.get(
+    `${process.env.NEXT_PUBLIC_API_URL}/healthcheck`,
+  );
+  return response.data;
+};
+
+export const isAxiosError451 = (error: any): boolean => {
```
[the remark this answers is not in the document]
gbarkhatov: Done

jeremy-babylonlabs (approved):

src/app/components/Connect/ConnectSmall.tsx
```diff
@@ -34,6 +37,27 @@ export const ConnectSmall: React.FC<ConnectSmallProps> = ({
...
+  // why the user might not be able to connect the wallet
+  const renderApiNotAvailableTooltip = () => {
+    if (!isBlocked && isApiNormal) return null;
+
+    return (
+      <>
+        <span
+          className="cursor-pointer text-xs"
+          data-tooltip-id="tooltip-connect"
+          data-tooltip-content={apiMessage}
```
vitsalis, line 52: I recommend if the error code is `ERROR_GEO_BLOCKED`, then we should always show "The Bitcoin Staking functionality is not available in your region.", regardless of what the error message says. This error code should only be used when someone is geo blocked. If we want to have another error, we can have another code for it.

For example, we could have a general purpose error code, in which case we display the error contained in the message. In the future, if we extend to include more error codes, I expect that we are going to have a big `if/else if/else`, in which on the `if/else if` we check for particular error codes and display custom messages for them and on the `else` we display the error message returned by the API response.

src/app/hooks/useHealthCheck.ts
```diff
@@ -0,0 +1,48 @@
...
+    }
+  }, [isError, error, showError, refetch]);
+
+  const isApiNormal = data?.status === HealthCheckStatus.Normal;
+  const isBlocked = data ? isGeoBlocked(data) : false;
+  const apiMessage = data?.message;
+
+  return {
+    isApiNormal,
+    isBlocked,
```
vitsalis, line 42:
```suggestion
    isGeoBlocked,
```

vitsalis (approved):
LGTM. Only a recommendation on displaying messages according to the error code and falling back to the error message as a default if the error code does not require custom support.

src/app/services/healthCheckService.ts
```diff
@@ -0,0 +1,40 @@
...
+        message: healthCheckAPIResponse.data,
+      };
+    } else {
+      throw new Error(API_ERROR_MESSAGE);
+    }
+  } catch (error: Error | any) {
+    if (isAxiosError451(error)) {
+      return {
+        status: HealthCheckStatus.GeoBlocked,
+        message: error.request?.response?.message || GEO_BLOCK_MESSAGE,
```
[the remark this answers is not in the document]
gbarkhatov: Done

src/app/services/healthCheckService.ts
```diff
@@ -0,0 +1,40 @@
...
+    const healthCheckAPIResponse = await fetchHealthCheck();
+    if (healthCheckAPIResponse.data) {
+      return {
+        status: HealthCheckStatus.Normal,
+        message: healthCheckAPIResponse.data,
+      };
+    } else {
+      throw new Error(API_ERROR_MESSAGE);
+    }
+  } catch (error: Error | any) {
```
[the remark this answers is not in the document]
gbarkhatov: Done

src/app/components/Header/Header.tsx
```diff
@@ -21,31 +21,29 @@ export const Header: React.FC<HeaderProps> = ({
...
-            <Logo />
-            <div className="flex flex-1">
-              <div className="hidden flex-1 xl:flex">
-                <TestingInfo />
-              </div>
+        <div className="container mx-auto flex w-full items-center justify-between gap-4 p-6 pb-4 md:pb-6">
+          <Logo />
+          <div className="flex flex-1">
+            <div className="hidden flex-1 xl:flex">
+              <TestingInfo />
```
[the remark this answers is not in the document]
gbarkhatov: Added

src/app/hooks/useHealthCheck.ts
gbarkhatov: `isGeoBlocked` is a funtion from `healthCheckService`, won't there be a confusion between this and hook return value? One is the boolean true/falce, another is a function that takes HealthCheckResult and returns a boolean

gbarkhatov: Done via:

- `isGeoBlocked` inside the service -> `isGeoBlockedResult` (this function takes the `HealthCheckResult` hence the name)
- `isBlocked` -> `isGeoBlocked` - just a boolean value

src/app/components/Connect/ConnectSmall.tsx
gbarkhatov: @vitsalis right now `errorCode` is `GATEWAY_GEO_BLOCKING`. I used the logic - if it's 451 stick with the `GEO_BLOCK_MESSAGE` and don't use `error?.response?.message` because of two things:
1. I am not sure we are 100% sure about the API contracts and all the responses/errors
2. In this particular case 451 means only one thing - geo blocking
3. All the future API refactorings can happen later on
4. I wouldn't want to add any more different additions to this PR, as it is already very different from what we usually do within the current system and approaches.

So in case of 451, instead of:

```ts
message: error.response?.message || GEO_BLOCK_MESSAGE
```

I am using

```ts
message: GEO_BLOCK_MESSAGE
```

In case of any other errors I will go with 

```ts
message: error.response?.message || error?.message || API_ERROR_MESSAGE
```

PR #27, opened 2024-08-02
feat: filter out utxos less or equal to 10k sats

jrwbabylonlab:


src/utils/utxo/index.ts
```diff
@@ -22,6 +24,9 @@ export const filterOrdinals = async (
   if (!utxos.length) {
     return [];
   }
+  // Filter UTXOs that has value less than 10k sats
+  utxos = filterLowValueUtxos(utxos);
```
vitsalis, line 28: Are we ok with modifying the same variable name with the filtered values?

vitsalis (approved):
just a minor question

PR #31, opened 2024-08-05
feat: add Github Actions

maiquanghiep:


Dockerfile
```diff
@@ -15,15 +15,20 @@ COPY tailwind.config.ts .
...
-# as next.js automatically replaces those during building
-# Later the docker-entrypoint.sh script finds such variables and replaces them
-# with the docker environment variables we have set
-RUN NEXT_PUBLIC_MEMPOOL_API=APP_NEXT_PUBLIC_MEMPOOL_API \
-    NEXT_PUBLIC_API_URL=APP_NEXT_PUBLIC_API_URL \
-    NEXT_PUBLIC_NETWORK=APP_NEXT_PUBLIC_NETWORK \
-    NEXT_PUBLIC_DISPLAY_TESTING_MESSAGES=APP_NEXT_PUBLIC_DISPLAY_TESTING_MESSAGES \
-    npm run build
+
+ARG NEXT_PUBLIC_MEMPOOL_API
```
vitsalis, line 19: Do we need to name the args the same way as the env variables? For example, why not just `MEMPOOL_API` for this one?

.github/workflows/publish.yaml
```diff
@@ -0,0 +1,122 @@
+name: docker_publish
+
+on:
+  push:
+    branches:
+      - 'hiep/add-github-actions'
+    # - 'main'
+    # - 'dev'
+    # tags:
+    # - '*'
```
vitsalis, line 10: Do we need those comments? What's the final form of this section?

.github/workflows/publish.yaml
```diff
@@ -0,0 +1,122 @@
...
+jobs:
+  lint_test:
+    uses: babylonlabs-io/.github/.github/workflows/reusable_node_lint_test.yml@hiep/add-nodejs-workflows
+  
+  docker_build:
+    needs: [lint_test]
+    runs-on: ubuntu-22.04
+    strategy:
+      matrix:
+        environment: [devnet, staging, mainnet]
```
vitsalis, line 21: What about testnet?

.github/workflows/publish.yaml
```diff
@@ -0,0 +1,122 @@
...
+        uses: actions/upload-artifact@v4
+        with:
+          name: simple-staking-${{ matrix.environment }}
+          path: /tmp/simple-staking-${{ matrix.environment }}.tar
+
+  dockerhub_publish:
+    runs-on: ubuntu-22.04
+    strategy:
+      matrix:
+        environment: [devnet, staging, mainnet]
```
vitsalis, line 57: Same here

.github/workflows/publish.yaml
```diff
@@ -0,0 +1,122 @@
...
+        if: startsWith(github.ref, 'refs/tags/')
+        run: |
+          docker tag simple-staking:${{ github.sha }} ${{ vars.DOCKERHUB_REGISTRY_ID }}/simple-staking:${{ github.ref_name }}-${{ matrix.environment }}
+          docker push ${{ vars.DOCKERHUB_REGISTRY_ID }}/simple-staking:${{ github.ref_name }}-${{ matrix.environment }}
+
+  ecr_publish:
+    runs-on: ubuntu-22.04
+    strategy:
+      matrix:
+        environment: [devnet, staging, mainnet]
```
vitsalis, line 90: Same here

Dockerfile
maiquanghiep: I used `NEXT_PUBLIC_*` for ARGS to match with the [.env.example](https://github.com/babylonlabs-io/simple-staking/blob/dev/.env.example) for consistency

maiquanghiep:
I will close this as it's better to merge to main

PR #41, opened 2024-08-06
feat: Add Github Actions

maiquanghiep:


.github/workflows/ci.yml
```diff
@@ -0,0 +1,13 @@
+name: ci
+
+on:
+  pull_request:
+    branches:
+    - '**'
+
+jobs:
+  lint_test:
+    uses: babylonlabs-io/.github/.github/workflows/reusable_node_lint_test.yml@hiep/add-nodejs-workflows
```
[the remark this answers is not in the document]
filippos47: For testing, the workflow is not merged to main yet.

.github/workflows/publish.yaml
filippos47, line 1: Ideally, this would become part of the reusable publish workflow.

filippos47 (commented):
Great effort!
Overall, a bit frustrating that we need to introduce a separate tagging convention just for a single service type (nodejs). Are there any alternatives that do not affect the tagging? @jrwbabylonlab 

This also breaks the automated devnet deployment for the simple-staking service on babylon-service-deployment:

https://github.com/babylonlabs-io/babylon-service-deployment/blob/330243364dd7f984e337bd00069b50c87423c5e2/.github/workflows/commit-schedule-checker.yml#L82

.github/workflows/ci.yml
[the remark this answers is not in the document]
filippos47: For testing, as the reusable workflow is not merged yet.

.github/workflows/publish.yaml
```diff
@@ -0,0 +1,124 @@
...
+    with:
+      run-build: true
+      run-unit-tests: true
+  
+  docker_build:
+    needs: [lint_test]
+    runs-on: ubuntu-22.04
+    strategy:
+      matrix:
+        environment: [devnet, staging, testnet, mainnet-private, mainnet]
```
vitsalis, line 23: This seems to go against the practice we want to achieve in which pipelines regarding our internal deployment environments are decoupled from the service repositories. Don't mind keeping it as is given that we're under time pressure, but let's track finding a solution for this.

.github/workflows/publish.yaml
```diff
@@ -0,0 +1,124 @@
...
+    environment: ${{ matrix.environment }}
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+
+      - name: Login to Docker Hub
```
vitsalis, line 32: Why do you need to login to docker hub to build an image?

vitsalis (approved):
lgtm, although this again introduces coupling between a service repository and an internal deployment. No problem with merging this now, but we should remove the tight coupling from here.

filippos47 (approved):
As we discussed offline, we can merge this for now.
The next iteration involves templatizing the extra publish logic into the reusable publish workflow.

.github/workflows/publish.yaml
maiquanghiep: Noted

.github/workflows/publish.yaml
maiquanghiep: Will be included in the next iteration, ideally the whole workflow should use the reusable publish workflow.

maiquanghiep:
@jrwbabylonlab Sorry my bad, selected squash but forgot to do the same after reloading the page

PR #44, opened 2024-08-06
bugfix - tooltip not displaying properly

### LINKED ISSUE #42: Tooltip not displaying properly on mobile phone

jeremy-babylonlabs:
closes #42

### DIFF
--- a/src/app/components/Modals/ConnectModal.tsx
+++ b/src/app/components/Modals/ConnectModal.tsx
@@ -224,7 +224,10 @@ export const ConnectModal: React.FC<ConnectModalProps> = ({
                           >
                             <AiOutlineInfoCircle />
                           </span>
-                          <Tooltip id={name} />
+                          <Tooltip
+                            id={name}
+                            className="max-w-xs whitespace-normal break-words"
+                          />
                         </div>
                       )}
                     </div>
--- a/src/app/components/Modals/ConnectModal.tsx
+++ b/src/app/components/Modals/ConnectModal.tsx
@@ -224,10 +224,7 @@ export const ConnectModal: React.FC<ConnectModalProps> = ({
                           >
                             <AiOutlineInfoCircle />
                           </span>
-                          <Tooltip
-                            id={name}
-                            className="max-w-xs whitespace-normal break-words"
-                          />
+                          <Tooltip id={name} className="tooltip-wrap" />
                         </div>
                       )}
                     </div>
--- a/src/app/components/Summary/Summary.tsx
+++ b/src/app/components/Summary/Summary.tsx
@@ -1,9 +1,16 @@
+import { useMemo, useState } from "react";
 import { AiOutlineInfoCircle } from "react-icons/ai";
 import { FaBitcoin } from "react-icons/fa";
 import { Tooltip } from "react-tooltip";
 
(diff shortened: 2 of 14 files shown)


jeremy-babylonlabs:
need a rerun on the build

gbarkhatov (approved):
1. Are there any places where this can happen? Besides `Stats`?
2. I thought we're hotfixing on main? @jrwbabylonlab @jeremy-babylonlabs. Don't forget to update the patch version

gbarkhatov (approved):
Don't forget to bump the patch

src/app/components/Summary/Summary.tsx
```diff
@@ -18,13 +27,41 @@ export const Summary: React.FC<SummaryProps> = ({
...
       <h3 className="mb-4 font-bold xl:mb-0">Your staking summary</h3>
       <div className="flex flex-1 justify-between gap-2">
         <div className="flex flex-col gap-1 text-sm xl:flex-1 xl:flex-row xl:items-center xl:justify-center xl:gap-2 xl:text-base">
-          <p className="dark:text-neutral-content">Total staked</p>
+          <div className="flex gap-1 items-center">
+            <p className="dark:text-neutral-content">Total staked</p>
+            <span
+              className="cursor-pointer text-xs"
+              data-tooltip-id="tooltip-total-staked"
+              data-tooltip-content={`Total staked is updated after ${paramWithCtx?.currentVersion?.confirmationDepth || 10} confirmations`}
```
vitsalis, line 58: Why are we adding this tool tip? This text is a bit confusing. Total staked is updated per bitcoin block and for a transaction to be included here it must have 10 btc confirmations. That's different from what's written.

I suggest we remove the tooltip altogether, I don't see a reason to add this after a feature freeze.

vitsalis (commented):
The additional tooltip should not be added while we're in a feature freeze. Apart from that lgtm.

src/app/components/Summary/Summary.tsx
vitsalis, line 58: Why are we adding this tool tip? This text is a bit confusing. Total staked is updated per bitcoin block and for a transaction to be included here it must have 10 btc confirmations. That's different from what's written.

I suggest we remove the tooltip altogether, I don't see a reason to add this after a feature freeze.

vitsalis (changes requested):

src/app/components/Summary/Summary.tsx
gbarkhatov: @jeremy-babylonlabs this part can be added to `dev`, not including this into `main`

src/app/components/Modals/ConnectModal.tsx
```diff
@@ -224,7 +224,10 @@ export const ConnectModal: React.FC<ConnectModalProps> = ({
                           >
                             <AiOutlineInfoCircle />
                           </span>
-                          <Tooltip id={name} />
+                          <Tooltip
+                            id={name}
+                            className="max-w-xs whitespace-normal break-words"
```
vitsalis, line 229: On all of them we added the class name `tooltip-wrap`, what's different about this tooltip?

vitsalis (approved):
lgtm, only minor comment

PR #53, opened 2024-08-07
fix: FAQ shall have dynamically injected confirmation depth

jrwbabylonlab:


src/app/components/FAQ/FAQ.tsx
```diff
@@ -1,17 +1,48 @@
...
 
 export const FAQ: React.FC<FAQProps> = () => {
+  const [paramWithCtx, setParamWithCtx] = useState<
+    ParamsWithContext | undefined
+  >();
   const { coinName } = getNetworkConfig();
+  const btcHeight = useBtcHeight();
+  const globalParams = useGlobalParams();
+
+  useEffect(() => {
```
jeremy-babylonlabs, line 24: this part of the code is actually reused many times, lets move it to a custom hook? can do it altogether in this PR or another PR

gbarkhatov (approved):
I don't like that the problematic part (context) is growing and growing. In this particular example we are mostly duplicating the thing we have in `Stats.tsx`

If we need things as they are right now - fast and good enough - we can go with this PR
If not - we should extract this as a hook and use in `FAQ.tsx` and `Stats.tsx`

```tsx
  useEffect(() => {
    if (!btcHeight || !globalParams.data) {
      return;
    }
    const paramsWithCtx = getCurrentGlobalParamsVersion(
      btcHeight + 1,
      globalParams.data,
    );
    if (!paramsWithCtx) {
      return;
    }
    setParamWithCtx(paramsWithCtx);
  }, [globalParams, btcHeight]);
```

jeremy-babylonlabs (approved):

PR #54, opened 2024-08-07
chore: improve the msg when staking amount exceeds the available balance

jrwbabylonlab:
image


jeremy-babylonlabs:
this sounds like a more casual tone, is that the brand tone we would go for? If so it looks good to me.

jeremy-babylonlabs (approved):

gbarkhatov (approved):

vitsalis (approved):

PR #66, opened 2024-08-12
fix: issue with fee UI react rendering loop

jrwbabylonlab:


src/utils/isStakingSignReady.ts
```diff
@@ -39,6 +40,11 @@ export const isStakingSignReady = (
       isReady: false,
       reason: "Please enter a valid staking period",
     };
+  } else if (stakingFeeSat == 0) {
+    return {
+      isReady: false,
+      reason: "Not enough funds to cover fees for staking",
```
vitsalis, line 46: Let's add a comment here as the `stakingFeeSat == 0` does not explain very well the situation behind the message.

vitsalis (approved):
LGTM. Let's have some documentation behind the decision.

jeremy-babylonlabs (approved):
lgtm, tested on chrome

src/utils/isStakingSignReady.ts
```diff
@@ -39,6 +40,11 @@ export const isStakingSignReady = (
       isReady: false,
       reason: "Please enter a valid staking period",
     };
+  } else if (stakingFeeSat == 0) {
```
gbarkhatov, line 43: Should we check for strict equals?

gbarkhatov (approved):

PR #78, opened 2024-08-15
Hotfix - clean input after staking and updated title text

jeremy-babylonlabs:
- [x] Improved staking clear input logic
- [x] updated staking and FP title text

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "simple-staking",
-  "version": "0.2.33",
+  "version": "0.2.34",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "simple-staking",
-      "version": "0.2.33",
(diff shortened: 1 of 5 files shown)


gbarkhatov (commented):
@jeremy-babylonlabs  Do we need to count the resets as state?

vitsalis:
I don't understand why are we counting resets

jeremy-babylonlabs:
using a reset counter can make sure the states rerenders in the lifecycle. but agreed and going with @jrwbabylonlab's approach for now.

gbarkhatov (approved):
@jeremy-babylonlabs I suggest we do add `strong` to steps, so it's more emphasized for the end user (for both steps)
```
      <p>
        <strong>Step-1:</strong> Select a finaltiy provider
      </p> 
```

Other than that PR does what it should - resets the amount

PR #85, opened 2024-08-16
fix: exclude the oneKey from injectable wallet

jrwbabylonlab:
OneKey still haven't release their changes until now (suppose to be monday this week or even earlier)
Also realise some users may already have oneKey installed and never upgrade it, then the issue will still exist for those users.

I decide to go with this super hacky and simple approach because don't want to mess up with the injectable logic in the last minute of launch.

image

also tested on mobile

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "simple-staking",
-  "version": "0.2.34",
+  "version": "0.2.35",
   "lockfileVersion": 3,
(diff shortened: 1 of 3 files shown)


gbarkhatov (approved):

jeremy-babylonlabs (approved):

jeremy-babylonlabs:
We should add an issue to keep track / revisit later

PR #103, opened 2024-08-21
feature: Adjust HW checkbox width

gbarkhatov:
Since we added `policy`, our first checkbox is wider than the third one: https://i.imgur.com/xmzXa1U.png. Looks bad
Changed the max width constraint: https://i.imgur.com/8b3uT4z.png


gbarkhatov:
@jrwbabylonlab @jeremy-babylonlabs this does not bump the patch

PR #105, opened 2024-08-22
chore: update mempool example url

jrwbabylonlab:


.env.example
```diff
@@ -1,4 +1,4 @@
-NEXT_PUBLIC_MEMPOOL_API=https://babylon.mempool.space
-NEXT_PUBLIC_API_URL=https://staking-api.testnet.babylonchain.io
+NEXT_PUBLIC_MEMPOOL_API=https://mempool.space
+NEXT_PUBLIC_API_URL=https://staking-api.staging.babylonchain.io
```
filippos47, line 2: How about keeping this to `testnet`?
We don't have any public information about the staging env, and it might be decommissioned at any point.

filippos47 (approved):

PR #113, opened 2024-08-26
fix: prefix FP url with https if no protocol present

jrwbabylonlab:
image


vitsalis:
Recommend that finality providers provide proper information instead of us cleaning invalid information they provided. In the future, we'll look out to not merge such PRs.

For me, this change just complicates the codebase. I would be much happier with just not displaying the URL if it is not a proper one.

gbarkhatov (approved):

PR #118, opened 2024-08-28
feature: Cookies toast

### LINKED ISSUE #107: Move privacy terms, terms condition & cookie should load in its own url
The terms and condition, cookie, as well as the privacy terms are currently shown in a modal.

We would like to host those files in a their own routes/url so that it can be linked from various places/sites.
For example, under the `http://babylonlabs.io/terms-of-use and http://babylonlabs.io/privacy-notice`.
Due to limited dev bandwidth, we can start with plain text and evolve in the future

gbarkhatov:
Adds Toast component, that can be customized in the future. Current use - cookies:
- https://i.imgur.com/OSZRWs7.png
- https://i.imgur.com/HFm6p5x.png
Closes #107

### DIFF
--- a/src/app/components/Toast/Toast.tsx
+++ b/src/app/components/Toast/Toast.tsx
@@ -26,7 +26,7 @@ export const Toast: React.FC<ToastProps> = ({
 
   return (
     <div className="toast toast-end z-10">
-      <div className="alert max-w-[20rem] flex flex-col">
+      <div className="alert sm:max-w-[20rem] flex flex-col">
         <p className="text-wrap text-xs">
           {message}{" "}
           {linkText && linkUrl && (
--- a/src/app/components/Toast/Toast.tsx
+++ b/src/app/components/Toast/Toast.tsx
@@ -1,0 +1,49 @@
+import { useEffect, useState } from "react";
+
+interface ToastProps {
+  message: string;
+  linkText?: string;
+  linkUrl?: string;
+  closeText?: string;
+  onClose: () => void;
+}
+
+export const Toast: React.FC<ToastProps> = ({
+  message,
+  linkText,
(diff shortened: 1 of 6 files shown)


src/app/components/Toast/Toast.tsx
```diff
@@ -0,0 +1,49 @@
...
+  // useEffect only runs on the client, so now we can safely show the UI
+  useEffect(() => {
+    setMounted(true);
+  }, []);
+
+  if (!mounted) return null;
+
+  return (
+    <div className="toast toast-end z-10">
+      <div className="alert max-w-[20rem] flex flex-col">
```
jeremy-babylonlabs, line 29: im thinking if it would be better to make it full width when its on mobile view

src/app/components/Toast/Toast.tsx
```diff
@@ -0,0 +1,49 @@
...
+              href={linkUrl}
+              target="_blank"
+              rel="noopener noreferrer"
+              className="font-bold underline"
+            >
+              {linkText}
+            </a>
+          )}
+        </p>
+        <button onClick={onClose} className="btn btn-primary btn-sm w-full">
```
jeremy-babylonlabs, line 43: Should the user be able to decline?

gbarkhatov: Right now there's no option to decline since it will be required to switch from the local storage to in-memory solution

src/app/components/Toast/Toast.tsx
gbarkhatov: Done

gbarkhatov:
@jrwbabylonlab converted to draft

gbarkhatov:
Closing right now until the decision is made

PR #123, opened 2024-09-05
fix: unable to copy FP public key if it's not a registered one

totraev:


gbarkhatov (approved):

gbarkhatov:
@jrwbabylonlab I think this can go to main? Or do we use `dev` in this case?

PR #142, opened 2024-09-12
feat: ignore invalid url in FP

totraev:


src/utils/url.ts
```diff
@@ -0,0 +1,5 @@
+const URL_REGEX = /^https?:\/\//;
```
gbarkhatov, line 1: What if it's http?

gbarkhatov (approved):

PR #148, opened 2024-09-15
use sentry to capture errors

supertong:
This is setup against the local instance of sentry on my laptop.


supertong:
> Can we have the sentry instance url passed in via config? Could we also use this config to control whether sentry error events will be abled? i.e if the sentry url is not passed in from env, we will not trigger the code execution of the error emitting.

Yes, we can provide that from the environment variable just like other `NEXT_PUBLIC_*`. I have updated the code to do that. And when the value is not provided, the Sentry will not send the event.

> For each error we emit, could we attach the env tag? for example, i would like to know if the error is coming from which btc network(signet/mainnet) and which backend api url the FE is pointing at.

Yes, also done that in the latest change. I have included both `environment` and `releases` in the config. If we want to include extra context to the event which is also possible but that has to be done inside the code.
image


> while you working on this ticket and going through the error handling logic, do you have any recommendations on how we shall improve the error handling in general?

I can see that we already have `ErrorContext` that used to handle expected errors. We don't have anything to handle unexpected errors before this PR. It's usually achieved through ErrorBoundary in React and NextJs allows you to do it through a `error.ts` file for each route segment. Since we only have one route for this app, the `global-error.tsx` I added in this PR should be enough to cover it. However, it's very simple and only offer user to refresh the app at the moment. I would suggest to add in more guidance like reaching out to your social channel and ask for help etc. I would also increase the event severity level in this case.

totraev (approved):

gbarkhatov:
@supertong can you please provide some instructions to test it out locally properly

supertong:
Hey guys, I recorded a video and hopefully that explains how to test this PR locally.
https://www.loom.com/share/7a6aa76245c94bcabe77f009c09249b0?sid=7c29f295-a46a-422c-808e-07e9fba3db6d
And the doc mentioned in the video to setup Sentry locally: https://develop.sentry.dev/self-hosted/

PR #154, opened 2024-09-17
add numeric format

jeremy-babylonlabs:


package.json
```diff
@@ -25,6 +25,7 @@
     "node": "22.3.0"
   },
   "dependencies": {
+    "@babylonlabs-io/btc-staking-ts": "0.3.0-canary.10",
```
gbarkhatov, line 28: is ordering happens automatically?

gbarkhatov (approved):

package.json
jeremy-babylonlabs: yes

Issue #164, opened 2024-09-19
E2E different addresses support


gbarkhatov:
E2E tests should support different addresses (Taproot, Native SegWit)

gbarkhatov:
Closing these tickets as they are not up to date anymore with the approach we took

Issue #136, opened 2024-09-09
E2E Text constants


gbarkhatov:
We use some inline text: `I certify that I have read`, `I acknowledge that Keystone via QR code`, `Connect to BTC network`,  or even `Staking Dashboard` as the page name.
We need to extract these as constants and use both in:
- our app
- e2e tests

So these 2 consume the stuff from one place

```mermaid
graph TD
    A[Text Constants] --> B[App]
    A[Text Constants] --> C[E2E Tests]
```

gbarkhatov:
@vitsalis do we want our app to consume text constants or do we just want to extract needed text to e2e text constants?

i.e. this:

```mermaid
graph TD
    A[Text Constants] --> B[App]
    A[Text Constants] --> C[E2E Tests]
```

or this

```mermaid
graph TD
    A[App]
    B[Text Constants] --> C[E2E Tests]
```

2nd approach is related to E2E only
1st approach will involve extraction of the needed (by E2E tests) labels to a file, then both app and e2e will consume the stuff from one place. Other application labels/texts (not touched by e2e) will not be extracted

vitsalis:
I think the first option will be much easier to maintain, as if we update the text, we can be confident that both the app and e2e tests will receive the updates. Might make the code a bit more unreadable though as the text will not be directly visible.

gbarkhatov:
Partially implemented in #193, but since we have other priorities full implementation will be when I have time (and #193 is merged)

gbarkhatov:
Closing these tickets as they are not up to date anymore with the approach we took

Issue #134, opened 2024-09-09
E2E real signed transactions


gbarkhatov:
In the future, we could automatically create a signed transaction based on the utxos, address, and public key we have defined above. This would allow us to potentially change any of those without having to regenerate a valid looking psbt.

This should affect `signPsbt` and `pushTx` methods

_Originally posted by @vitsalis in https://github.com/babylonlabs-io/simple-staking/pull/77#discussion_r1745299615_

Additional: https://github.com/babylonlabs-io/simple-staking/pull/77#discussion_r1745300437

gbarkhatov:
@vitsalis @jrwbabylonlab 
Since playwrights node and browser environments are different, and we need our wallet to work with bitcoinjs (so as other deps, like BIP32 or BIP39), we will need to bundle the wallet into the JS file:

```mermaid
graph TD
    A[Webpack bundles wallet] --> B[Wallet is bundled as JS file]
    B --> C[Wallet JS file can be consumed]
    C --> D[Wallet JS file is injected into Playwright browser]
    D --> E[Playwright browser executes e2e tests with wallet]
```

gbarkhatov:
Closing these tickets as they are not up to date anymore with the approach we took

Issue #132, opened 2024-09-09
E2E page.evaluate outside imports / classes


gbarkhatov:
We can have inline mocked sample wallet 
```
await page.evaluate(() => {
  const btcWallet = {
    connectWallet: () => {
      return btcWallet
    }
  }
})
```

Improvements:
- import the wallet implementation from the outside file so it's easier to manage
- wallet to be a `Class` implementing `WalletProvider`
- wallet being able to call outside methods (mempool api?)
- constants to be imported/exported

All of this should be injected to window.btcwallet on our E2E tests

gbarkhatov:
Closing these tickets as they are not up to date anymore with the approach we took

PR #172, opened 2024-09-25
feature: E2E real data

gbarkhatov:
Added real wallet E2E tests

1. Wallets are generated via real mnemonic. Both `mainnet/signet` and `taproot/native segwit`
2. Since `window.btcwallet` is something we need to inject, and we do want our transactions to be real, we have to build our wallets with all the necessary bitcoin libs. Currently, I use `webpack` for that
3. The only `mocking` part is the initial `UTXOs` and `POST` requests. I.e. under the hood real BTC signatures are happening
4. By default, Playwright will use the `.env.local` that you use. If you want to use a specific `mainnet` and `signet` env variables, you need to create `.env.mainnet` and `.env.signet`.
5. You can use whatever wallets you want in your tests, but some of them require the specific ones. For example, in case of `balance and address` test, we can connect both `signet` and `mainnet` wallets and check the stuff. But anything with the `signing` involved will require specific env variables. You cannot use `mainnet` env variables with global parameters expecting `X` `mainnet` BTC Tip height on `signet` wallet. Since `signet` tip height is going to be different` So `staking`, `unbonding`, `withdrawal` use a set of mapping and test required network wallets only. That's because in our app we use `src/app/context/mempool/BtcHeightProvider.tsx` which will give `X` while wallet mempool endpoint might give `Y` depending on the `env` variables.

Besides reviewing the code I expect you run the instructions in `Readme.md`:
- `npm i` <- install the dependencies
- `npx playwright install` <- this is important
- `npm run build:btcwallet` <- to build both wallets
- `npm run test:e2e` <- for your `.env.local` or
- `npm run test:e2e:full` <- for your `.env.mainnet` and `.env.signet`

Added an E2E section in Readme.md

Video: https://youtu.be/jppgJa9L6gM

Closes:
- #124 
- #132 
- #134 
- #164 

Partly closes, so the PR is not even bigger:
- #136


e2e/balanceAddress.spec.ts
```diff
@@ -1,20 +1,90 @@
...
+  nativeSegwit: {
+    mainnet: nativeSegwitMainnetBalance,
+    signet: nativeSegwitSignetBalance,
+  },
+};
+
+// Reusable function for checking address
+async function checkAddress(page: Page, expectedAddress: string) {
+  const address = await page.getByTestId("address").textContent();
+  expect(address).toBe(trim(expectedAddress));
```
totraev, line 57: I think it's better to remove `expect` condition from here and rename this helper function to regular getter. First of all this is more declarative way of writing tests + this function violates the single responsibility principle (it's a getter and expect function at the same time)

e2e/helper/connect.ts
totraev, line 1: Maybe we should wrap up all these helper functions into [fixtures](https://playwright.dev/docs/test-fixtures)? Is there any example of using `helpers` in there docs? I think that fixtures is more PlayWright way

package.json
```diff
@@ -16,27 +16,32 @@
     "sort-imports": "eslint --fix .",
     "test": "jest",
     "test:watch": "jest --watch",
+    "build:btcwallet:mainnet": "webpack --config webpack.btcWallet.config.js --env NEXT_PUBLIC_NETWORK=mainnet",
```
totraev, line 19: This looks weird. Instead of putting everything into our DApp application we could create a separate repo for test wallet and deploy it to npm. We could create a very simple browser extension for test purposes and inject it as devDependency into our main application.

I like the Tomo way of doing thing, they have created a separate repo even for their (our) wallet [providers](https://github.com/UnyxTech/tomo-wallet-provider), what makes much easier to maintain their SDK and Wallet Providers

gbarkhatov: Yep, as I mentioned on the call, we now have `simple-staking` and a lot of stuff to build the wallet for e2e purposes.
Question though is are we ready to extract wallet providers and all of that outside of simple-staking? Is it a good time **now** or should we do it later?

totraev: All this stuff around wallet providers is so confusing for me right now... If we have mobile/injectable provider in our app why just not use it for the test wallet? We can implement this interface in a separate repo and manually inject it into global window object before we run tests

totraev: If I'm getting things right here is the list of packages and modules which we need extract to a new repo

`import "core-js/web/url";

import * as ecc from "@bitcoin-js/tiny-secp256k1-asmjs";
import * as bip32 from "@scure/bip32";
import * as bip39 from "@scure/bip39";
import * as bitcoin from "bitcoinjs-lib";
import ECPairFactory from "ecpair";

import { getNetworkFees, getTipHeight } from "@/utils/mempool_api";
import { getPublicKeyNoCoord, isTaproot } from "@/utils/wallet";
import {
  Fees,
  InscriptionIdentifier,
  Network,
  UTXO,
  WalletProvider,
} from "@/utils/wallet/wallet_provider";

import { nativeSegwitMainnetUTXOs } from "../mock/mainnet/nativeSegwit/utxos";
import { taprootMainnetUTXOs } from "../mock/mainnet/taproot/utxos";
import { nativeSegwitSignetUTXOs } from "../mock/signet/nativeSegwit/utxos";
import { taprootSignetUTXOs } from "../mock/signet/taproot/utxos";`

gbarkhatov: > If we have mobile/injectable provider in our app why just not use it for the test wallet
I did not find the solution for the wallet to be "as it is" in typescript code and we are just injecting the stuff into window.btcwallet that works in node.js and Chrome (other browsers) Playwright env. As soon as we are starting to say "bitcoinjs sign this" or "bip39/32 get the private key" or "mempool fetch the data" it will cause problems
The only solution to this issue that I found is to build a JS wallets packages (build) that have everything they need. Then playwright injects it and calls needed methods

Issue #124, opened 2024-09-06
E2E Able to run test in testnet


jrwbabylonlab:
In our very first e2e test PR, we have added the tests to run against with the network of `mainnet`. but it would be ideal to have the tests automation also run against testnets(signet)
For reference: https://github.com/babylonlabs-io/simple-staking/pull/77

gbarkhatov:
> hi @gbarkhatov is this done?

Once pr is simplified and merged, but the change is in the pr

gbarkhatov:
Closing these tickets as they are not up to date anymore with the approach we took

PR #193, opened 2024-10-02
feature: E2E real wallet data using esbuild

gbarkhatov:
Added real wallet E2E tests

1. Wallets are generated via real mnemonic. Both `mainnet/signet` and `taproot/native segwit`
2. Since `window.btcwallet` is something we need to inject, and we do want our transactions to be real, we have to build our wallets with all the necessary bitcoin libs. We do it using `esbuild`
3. The only `mocking` part is the initial `UTXOs` and `POST` requests. I.e. under the hood real BTC signatures are happening
4. By default, Playwright will use the `.env.local` that you use. If you want to use a specific `mainnet` and `signet` env variables, you need to create `.env.mainnet` and `.env.signet`.
5. You can use whatever wallets you want in your tests, but some of them require the specific ones. For example, in case of `balance and address` test, we can connect both `signet` and `mainnet` wallets and check the stuff. But anything with the `signing` involved will require specific env variables. You cannot use `mainnet` env variables with global parameters expecting `X` `mainnet` BTC Tip height on `signet` wallet. Since `signet` tip height is going to be different` So `staking`, `unbonding`, `withdrawal` use a set of mapping and test required network wallets only. That's because in our app we use `src/app/context/mempool/BtcHeightProvider.tsx` which will give `X` while wallet mempool endpoint might give `Y` depending on the `env` variables.

Besides reviewing the code I expect you run the instructions in `Readme.md`:
- `npm i` <- install the dependencies
- `npx playwright install` <- this is important
- `npm run test:e2e` <- for your `.env.local` or
- `npm run test:e2e:full` <- for your `.env.mainnet` and `.env.signet`

Added an E2E section in Readme.md

Video: https://youtu.be/jppgJa9L6gM

Closes:
- #124 
- #132 
- #134 
- #164 

Partly closes, so the PR is not even bigger:
- #136


gbarkhatov:
@totraev @jrwbabylonlab @jeremy-babylonlabs network is mocked, tests are working with wifi turned off

gbarkhatov:
> could you set up the playwright config file that can help run the tests a few times? maybe 2-5 times is enough
This might be slow 😅

totraev (changes requested):
In general:

1. I think we should follow community standards and use [fixtures](https://playwright.dev/docs/test-fixtures) over helper functions.
2. If we don't plan to use a separate repo for test wallet we need optimise current implementation. At the moment we recompile and rebuild wallet before each test(!):
  `test.beforeEach` > `setupWalletConnection` > `injectBTCWallet` > `build`
3. Instead of using `interceptRequest` I'd suggest to run mock [server](https://github.com/typicode/json-server) and keep all mock there. It'll significantly simplify things

e2e/balanceAddress.spec.ts
```diff
@@ -1,20 +1,96 @@
...
+const expectedBalances: Record<AddressType, Record<Network, number>> = {
+  taproot: {
+    mainnet: taprootMainnetBalance,
+    signet: taprootSignetBalance,
+  },
+  nativeSegwit: {
+    mainnet: nativeSegwitMainnetBalance,
+    signet: nativeSegwitSignetBalance,
+  },
+};
```
totraev, line 53: We can move this logic together with current `network` to constants folder and return only taproot/nativeSegwit values depends on current `network` value:

```ts
const addresses = {
  mainnet: {
    taproot: 'address',
    nativeSegwit: 'address',
  },
  signet: {
    taproot: 'address',
    nativeSegwit: 'address',
  }
}[network];
```

e2e/balanceAddress.spec.ts
```diff
@@ -1,20 +1,96 @@
...
-    expect(balance).toBe("0.12345678 BTC");
-  });
-
-  test("address is correct", async ({ page }) => {
-    const address = await page.getByTestId("address").textContent();
-    expect(address).toBe("bc1p...97sd");
-  });
+  // Iterate over each address type
+  for (const type of addressTypes) {
+    test.describe(`${type} address on ${network}`, () => {
```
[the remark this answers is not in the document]
totraev: @jrwbabylonlab why we need test against multiple networks if everything is mocked?

e2e/balanceAddress.spec.ts
```diff
@@ -1,20 +1,96 @@
...
+      }) => {
+        const expectedAddress = expectedAddresses[type][network];
+        const expectedBalance = expectedBalances[type][network];
+
+        // Perform address check
+        await checkAddress(page, expectedAddress);
+
+        // Perform balance check
+        await checkBalance(page, expectedBalance);
+      });
```
totraev, line 93: Test is pretty small, no need in additional wrapper. Weird to see test without `expect` expression 

```suggestion
      test(`should verify the ${type} address and balance on ${network}`, async ({
        page,
      }) => {
        const expectedAddress = expectedAddresses[type][network];
        const expectedBalance = expectedBalances[type][network];

        // Perform address check
         const address = await page.getByTestId("address").textContent();
         expect(address).toBe(trim(expectedAddress));

        // Perform balance check
        const balanceText = await page.getByTestId("balance").textContent();
        const balance = maxDecimals(extractNumericBalance(balanceText), 8);
         expect(balance).toBe(satoshiToBtc(expectedBalance));
      });
```

e2e/helper/injectBTCWallet.ts
```diff
@@ -1,47 +1,70 @@
...
+
+export const injectBTCWallet = async (
+  page: Page,
+  network: string = "mainnet",
+  walletType: string = "taproot",
+) => {
+  try {
+    const walletPath = path.resolve(__dirname, "btcWallet.ts");
+
+    const result = await build({
```
totraev, line 17: I still think we should move it to a separate repo

e2e/helper/mockNetwork.ts
```diff
@@ -0,0 +1,50 @@
...
+    "**/v1/stats**",
+    200,
+    isOnMainnet ? mainnetStats : signetStats,
+  );
+  await interceptRequest(
+    page,
+    "**/tip/height**",
+    200,
+    isOnMainnet ? mainnetTipHeight : signetTipHeight,
+  );
```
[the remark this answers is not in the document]
totraev: @jrwbabylonlab mock server is a better option

gbarkhatov:
Suggestion - finalize after the Tomo dual wallet onboarded

gbarkhatov:
One thing we discussed with @jrwbabylonlab - we can skip the tests because of the UI, wallet, processes (staking, unbonding, withdrawal) and enable them as soon as the needed parts are working and settled. I.e. when we add working wallet solution we can enable wallet e2e tests

gbarkhatov:
Converting this to draft until new changes are implemented related to Tomo wallet injectable and selectors

gbarkhatov:
@jrwbabylonlab @totraev @jeremy-babylonlabs I suggest closing this PR as it is mostly irrelevant now for a while

PR #202, opened 2024-10-04
use correct height for unbonding model (#196)

jeremy-babylonlabs:
* fix: improvement on using the correct height for unbonding model ---------

added `getVersionInfoByHeight` in `versionInfoContext`


src/app/context/api/VersionInfo.tsx
```diff
@@ -43,13 +45,22 @@ export function VersionInfoProvider({ children }: PropsWithChildren) {
...
         isLoading: isVersionLoading || isHeightLoading,
+        getVersionInfoByHeight: () => undefined,
       };
 
+    const currentVersionInfo = getCurrentGlobalParamsVersion(height + 1, versions);
+
+    const getVersionInfoByHeight = (targetHeight: number) => {
+      const { currentVersion } = getCurrentGlobalParamsVersion(targetHeight, versions);
+      return currentVersion;
+    };
```
totraev, line 56: You don't need to put this logic here. Instead you can extend it by using a custom hook like this one:

```ts
export function useVersionByHeight(height: number) {
  const { versions } = useVersionInfo()
  
  return useMemo(() => getCurrentGlobalParamsVersion(height, versions), [versions, height])
}
```

src/app/components/Modals/UnbondWithdrawModal.tsx
```diff
@@ -15,26 +16,37 @@ export const MODE_WITHDRAW = "withdraw";
...
   const { coinName, networkName } = getNetworkConfig();
   const versionInfo = useVersionInfo();
 
-  const globalParams = versionInfo?.currentVersion;
+  const delegation = delegationsAPI.find(
+    (delegation) => delegation.stakingTxHashHex === txID,
+  );
+  if (!delegation) {
+    throw new Error("Delegation not found");
+  }
```
totraev, line 45: It's not good idea to throw errors inside render method. Do we already have error boundaries?

src/app/components/Modals/UnbondWithdrawModal.tsx
```diff
@@ -15,26 +16,37 @@ export const MODE_WITHDRAW = "withdraw";
...
+  delegationsAPI,
+  txID,
 }) => {
   const { coinName, networkName } = getNetworkConfig();
   const versionInfo = useVersionInfo();
 
-  const globalParams = versionInfo?.currentVersion;
+  const delegation = delegationsAPI.find(
+    (delegation) => delegation.stakingTxHashHex === txID,
+  );
```
totraev, line 42: Wrap up any computation logic in useMemo inside render method
```suggestion
  const delegation = useMemo(() => delegationsAPI.find(
    (delegation) => delegation.stakingTxHashHex === txID,
  ), [delegationsAPI, txID]);
```

src/app/components/Modals/UnbondWithdrawModal.tsx
```diff
@@ -15,26 +16,37 @@ export const MODE_WITHDRAW = "withdraw";
...
+  if (!delegation) {
+    throw new Error("Delegation not found");
+  }
+  
+  const globalParams = versionInfo?.getVersionInfoByHeight(
+    delegation.stakingTx.startHeight,
+  );
```
totraev, line 49: Use `useVersionByHeight` instead

src/app/components/Modals/UnbondWithdrawModal.tsx
jeremy-babylonlabs: yep you're right we don't have error boundaries yet. I have removed this. for now the display will be '-' if delegation isnt found.

src/app/context/api/VersionInfo.tsx
```diff
@@ -24,6 +24,14 @@ export function useVersionInfo() {
...
 }
 
+export function useVersionByHeight(height: number) {
+  const { data: versions } = useVersions();
+  return useMemo(
+    () => getCurrentGlobalParamsVersion(height, versions ?? []),
+    [versions, height],
+  );
+}
+
```
totraev, line 34: PLs move it to our `hooks` folder @jeremy-babylonlabs

totraev (approved):

gbarkhatov (approved):

jeremy-babylonlabs:
this is for syncing into dev

PR #224, opened 2024-10-12
Add error pages

supertong:
This PR adds in error pages.

* error.tsx - for uncaught errors happens in the page
* global-error.tsx - for uncaught errors that happens in layers above page, for example Providers.
* not-found.tsx - for 404 route not found

I didn't change the overall design to make sure the error page styles are consistent with the existing pages. The design from Figma seems to be the new UI design?

Also, I haven't added any actions to those two buttons as I need information on where they should redirect user to or which email should they send to.

I will also post a recording video here.
https://www.loom.com/share/b18ea0ec70a94ccfb77ca4db480f020c?sid=7a743d20-5684-4dc1-bc1f-5a54d2742061


src/app/layout.tsx
```diff
@@ -20,23 +23,22 @@ export default function RootLayout({
...
-      <meta property="og:image" content={`${getNetworkAppUrl()}/og.png`} />
-      <meta name="twitter:card" content="summary_large_image" />
-      <meta name="twitter:title" content="BTC Staking Dashboard" />
-      <meta name="twitter:description" content="BTC Staking Dashboard" />
-      <meta name="twitter:image" content={`${getNetworkAppUrl()}/og.png`} />
-      <meta name="twitter:image:type" content="image/png" />
-      <meta name="twitter:image:width" content="2048" />
-      <meta name="twitter:image:height" content="1170" />
+      <head>
+        <MetaTags />
```
totraev, line 27: I think we can use next.js [default ](https://nextjs.org/docs/app/building-your-application/optimizing/metadata)approach for metadata instead of this

totraev (approved):

gbarkhatov (approved):

src/app/layout.tsx
supertong: Thanks for pointing that out. It would definitely work better for `layout.tsx` but I am afraid the `global-error.tsx` would have to do it this way though.

supertong: I had to resolve this conversation to get the PR merge in but feel free to continue the discussion if you want to. :)

PR #251, opened 2024-10-23
feature: Ordinals filtering modal

### LINKED ISSUE #207: Offer staker option to filter ordinals
Instead of always attempting to filter staker ordinals automatically, we should offer a different approach where stakers can choose whether to include or exclude UTXOs with ordinals.

The process would work as follows:
- The user connects their wallet, and we check against the ordinal service.
- If ordinals are found, we display a message asking the user to confirm whether they want to include the UTXO or filter it out from the staking balance.
- By default (in case backend never respond or returned error) we exclude the UTXOs with ordinals
       

https://www.figma.com/design/EbglGlHvz6C8AZpzijQFc6/Sandbox?node-id=46-14538&t=0BAQJe1GP9UeR0CN-1 

The text of the checkbox at wallet connect modal shall be updated. please ask legal team about the exact wordings

**NOTE: You should raise PR against main**

gbarkhatov:
Adds a modal and settings option to filter out ordinals

Closes #207

### DIFF
--- a/src/app/components/Connect/ConnectSmall.tsx
+++ b/src/app/components/Connect/ConnectSmall.tsx
@@ -20,6 +20,8 @@ interface ConnectSmallProps {
   address: string;
   btcWalletBalanceSat?: number;
   onDisconnect: () => void;
+  shouldFilterOrdinals: boolean;
+  setShouldFilterOrdinals: (value: boolean) => void;
 }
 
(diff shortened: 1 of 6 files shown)


jeremy-babylonlabs (approved):

totraev (approved):
LGTM

PR #258, opened 2024-10-25
feature: Resolve filter ordinals

gbarkhatov:
- `main` branch uses a combination of non-invasive changes to show the filter ordinals
- `dev` branch uses hooks

Whether or not to show the `FilterOrdinalsModal` is a part of `BTCWalletProvider`. Initially I went with the `BTCWalletProvider` handling the "show/no show" the modal and `shouldFilterOrdinals` to be a part of `AppState`. But then we cannot just put the `FilterOrdinalsModal` inside this provider, since it has to consume `AppState` `shouldFilterOrdinals` commit: af8ea0bf8eda03f3ec21e30fec0e9aad621d7487

But then I switched to even simpler solution - filtering ordinals (modal, should show modal, should filter or not) is just a part of BTCWalletProvider. Commit: e366a752447fe88b5a161163c8efae16078e8c09

This solution provides us an easier way to manage the stuff. Also, as soon as this modal is a part of Tomo wallet connect, we can easier get rid of the modal itself, so as we will get Tomos internal "should filter ordinals or not"

### DIFF
--- a/src/app/context/wallet/BTCWalletProvider.tsx
+++ b/src/app/context/wallet/BTCWalletProvider.tsx
@@ -47,6 +47,8 @@ interface BTCWalletContextProps {
   getUtxos: (address: string, amount?: number) => Promise<UTXO[]>;
   getBTCTipHeight: () => Promise<number>;
   getInscriptions: () => Promise<InscriptionIdentifier[]>;
+  filterOrdinalsModalOpen: boolean;
+  setFilterOrdinalsModalOpen: (open: boolean) => void;
 }
 
 const BTCWalletContext = createContext<BTCWalletContextProps>({
@@ -69,6 +71,8 @@ const BTCWalletContext = createContext<BTCWalletContextProps>({
   getUtxos: () => Promise.resolve([]),
   getBTCTipHeight: () => Promise.resolve(0),
   getInscriptions: () => Promise.resolve([]),
+  filterOrdinalsModalOpen: false,
+  setFilterOrdinalsModalOpen: () => {},
 });
 
 export const BTCWalletProvider = ({ children }: PropsWithChildren) => {
@@ -79,7 +83,7 @@ export const BTCWalletProvider = ({ children }: PropsWithChildren) => {
   const [address, setAddress] = useState("");
 
   const { showError } = useError();
-  const { open, disconnect, isConnected, providers } = useWalletConnection();
+  const { open, isConnected, providers } = useWalletConnection();
 
   const btcDisconnect = useCallback(() => {
     setBTCWalletProvider(undefined);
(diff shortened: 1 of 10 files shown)


src/app/hooks/api/useUTXOs.ts
```diff
@@ -6,14 +6,20 @@ import { filterOrdinals } from "@/utils/utxo";
...
 
     const mempoolUTXOs = await getUtxos(address);
+    // Return UTXOs without filtering if not required
+    if (!shouldFilterOrdinals) {
+      return mempoolUTXOs;
+    }
+
     const filteredUTXOs = await filterOrdinals(
       mempoolUTXOs,
       address,
```
totraev, line 25: useUTXOs should be only responsible for fetching data (not filtering it). `filderOrdinals` method is a legacy

src/app/context/wallet/BTCWalletProvider.tsx
```diff
@@ -88,6 +93,9 @@ export const BTCWalletProvider = ({ children }: PropsWithChildren) => {
     setAddress("");
   }, []);
 
+  const [filterOrdinalsModalOpen, setFilterOrdinalsModalOpen] = useState(false);
+  const [shouldFilterOrdinals, setShouldFilterOrdinals] = useState(true);
```
totraev, line 97: We need make wallet providers stateless, we moved all application's state to `/app/state` folder

totraev (changes requested):

src/app/context/wallet/BTCWalletProvider.tsx
```diff
@@ -221,6 +234,12 @@ export const BTCWalletProvider = ({ children }: PropsWithChildren) => {
   return (
     <BTCWalletContext.Provider value={btcContextValue}>
       {children}
+      <FilterOrdinalsModal
+        open={filterOrdinalsModalOpen}
+        onClose={setFilterOrdinalsModalOpen}
+        shouldFilterOrdinals={shouldFilterOrdinals}
+        setShouldFilterOrdinals={setShouldFilterOrdinals}
+      />
```
totraev, line 242: I don't think this modal relates to BTCWalletProvider. It can be placed anywhere inside our DApp

totraev (changes requested):

totraev:
If you want to place ordinal's logic inside `BTCWalletProvider` then you need modify `getUtxos` method cause currently it ignores your `shouldFilterOrdinals` flag and always return all UTXOs.

src/app/state/index.tsx
```diff
@@ -22,15 +26,22 @@ export interface AppState {
   firstActivationHeight: number;
   isError: boolean;
   isLoading: boolean;
+  ordinalsExcluded: boolean;
+  includeOrdinals: () => void;
+  excludeOrdinals: () => void;
```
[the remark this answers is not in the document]
totraev: @jrwbabylonlab Yeah, it'll, but for now it's okay. We can decompose this context provider in the future (for performance optimisation reasons)

totraev (approved):

src/app/state/index.tsx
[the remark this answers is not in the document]
gbarkhatov: @jrwbabylonlab @totraev 
https://github.com/babylonlabs-io/simple-staking/issues/265

gbarkhatov:
Additional:
https://github.com/babylonlabs-io/simple-staking/issues/266

PR #273, opened 2024-10-31
fix: upgrade and adjust the tomo connect to pave way for broadcast BB…

jrwbabylonlab:
Registry require the proto ts to be upgraded to produce ts types instead of methods.
Alternately, we can generate in this repo for now. but i will leave it for a seperate PR

https://github.com/babylonlabs-io/babylon-proto-ts/issues/14

### DIFF
--- a/src/app/context/wallet/CosmosWalletProvider.tsx
+++ b/src/app/context/wallet/CosmosWalletProvider.tsx
@@ -24,7 +24,7 @@ interface CosmosWalletContextProps {
 }
 
 const getSigningStargateClientDefault = async () => {
-  throw new Error("Not implemented");
+  throw new Error("Not initialized");
 };
 
 const CosmosWalletContext = createContext<CosmosWalletContextProps>({
--- a/package-lock.json
+++ b/package-lock.json
@@ -11,6 +11,8 @@
         "@babylonlabs-io/btc-staking-ts": "0.3.0",
         "@bitcoin-js/tiny-secp256k1-asmjs": "2.2.3",
         "@bitcoinerlab/secp256k1": "^1.1.1",
+        "@cosmjs/proto-signing": "^0.32.4",
+        "@cosmjs/stargate": "^0.32.4",
         "@keystonehq/animated-qr": "^0.8.6",
         "@keystonehq/keystone-sdk": "^0.4.1",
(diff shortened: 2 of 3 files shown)


totraev (approved):

src/app/context/wallet/CosmosWalletProvider.tsx
```diff
@@ -10,56 +11,53 @@ import {
...
   }, []);
 
   const connectCosmos = useCallback(async () => {
     if (!providers.cosmosProvider) return;
 
     try {
-      const chainID = providers.cosmosProvider.getChainId();
-      const cosmosInfo =
-        await providers.cosmosProvider.provider.getKey(chainID);
+      await providers.cosmosProvider.connectWallet();
```
totraev, line 54: @jrwbabylonlab why we need this method here?

totraev (approved):

src/app/context/wallet/CosmosWalletProvider.tsx
```diff
@@ -1,3 +1,4 @@
+import { SigningStargateClient } from "@cosmjs/stargate";
```
gbarkhatov, line 1: We are importing this, but it is not a direct dependency of the app (we don't have `@cosmjs/stargate` in the package.json)

src/utils/wallet/bbnRegistry.ts
```diff
@@ -0,0 +1,6 @@
+import { Registry } from "@cosmjs/proto-signing";
```
gbarkhatov, line 1: Same here, not a direct dependency

src/app/context/wallet/CosmosWalletProvider.tsx
```diff
@@ -10,56 +11,53 @@ import {
...
-  pubKey: string;
   connected: boolean;
   disconnect: () => void;
   open: () => void;
-  signArbitrary: (message: string) => Promise<any>;
+  getSigningStargateClient(): Promise<SigningStargateClient>;
 }
 
+const getSigningStargateClientDefault = async () => {
+  throw new Error("Not implemented");
```
gbarkhatov, line 27: Should we add a todo here?

gbarkhatov (approved):

PR #322, opened 2024-11-08
clean up punnycode warning

supertong:


package.json
```diff
@@ -90,5 +90,10 @@
     "ts-jest": "^29.1.4",
     "ts-node": "^10.9.2",
     "typescript": "^5"
+  },
+  "overrides": {
+    "whatwg-url@11.0.0": {
+      "tr46": "4.0.0"
```
supertong, line 96: It's requiring version 3.0.0 at the moment but 4.0.0 is just 2 commits away so I reckon it should be safe. In addition to that, it's used by `jest` so if anything breaks, it should break the test first. 🤞

gbarkhatov (approved):

supertong:
retarget to main.

PR #335, opened 2024-11-12
feature: Pending verification modal

gbarkhatov:
Addition of Pending Verification modal with the current design

To use it:

```tsx
  const [pendingVerificationOpen, setPendingVerificationOpen] = useState(false);
  const [verified, setVerified] = useState(false);
  const [awaitingWalletResponse, setAwaitingWalletResponse] = useState(false);
  const handlePendingOpen = async () => {
    // open the modal
    setPendingVerificationOpen(true);
    await new Promise((resolve) => setTimeout(resolve, 3000)); // simulate waiting for BBN verification
    setVerified(true);
  };
  const handlePendingClose = () => {
    // clean out the state
    setVerified(false);
    setAwaitingWalletResponse(false);
    setPendingVerificationOpen(false);
  };
  const handleStakeSignBroadcast = async () => {
    console.log("start staking");
    setAwaitingWalletResponse(true); // waiting for wallet response
    await new Promise((resolve) => setTimeout(resolve, 3000));
    setAwaitingWalletResponse(false);
    console.log("staking done");
    handlePendingClose();
  };
```

Video: https://youtu.be/6bIyE6O3vIk

Things that should be added to this modal:
1. We should poll for a delegation. If it's verified - change the verified flag. Back-end endpoint to be provided by @jeremy-babylonlabs 
2. Local storage. TX should be saved to local storage, to show that it's pending verification. Once verified - we can click the button (either in modal or in delegations table) to sign the staking tx and broadcast it to BTC

### DIFF
--- a/src/app/components/Modals/PendingVerificationModal.tsx
+++ b/src/app/components/Modals/PendingVerificationModal.tsx
@@ -1,0 +1,71 @@
+import { FaCheckCircle } from "react-icons/fa";
+
+import { getNetworkConfig } from "@/config/network.config";
+
+import { GeneralModal } from "./GeneralModal";
+
+interface PendingVerificationModalProps {
+  open: boolean;
+  onClose: (value: boolean) => void;
+  verified: boolean;
+  onStake?: () => void;
+  awaitingWalletResponse: boolean;
+}
+
+const Verified = () => (
+  <>
+    <FaCheckCircle className="text-5xl text-success" />
+    <h3 className="text-xl text-center">Verified</h3>
+    <p className="text-sm text-center">
+      Your request has been verified by the babylon blockchain. You can now
+      stake
+    </p>
+  </>
+);
+
+const NotVerified = () => (
+  <>
+    <span className="loading loading-spinner loading-lg text-primary" />
+    <h3 className="text-xl text-center">Pending Verification</h3>
+    <p className="text-sm text-center">
+      The babylon blockchain has received your request. Please wait while we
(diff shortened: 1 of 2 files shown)


src/app/components/Modals/PendingVerificationModal.tsx
```diff
@@ -0,0 +1,71 @@
...
+    </p>
+  </>
+);
+
+export function PendingVerificationModal({
+  open,
+  onClose,
+  verified,
+  onStake,
+  awaitingWalletResponse,
```
totraev, line 42: Naming: awaitingWalletResponse -> loading

totraev (approved):

src/app/components/Modals/PendingVerificationModal.tsx
gbarkhatov: @totraev but we use `awaitingWalletResponse` in such cases for the staking and unbonding/withdrawal
So `loading` in this particular modal is just the flag. While not `verified` it will be in loading state
But `awaitingWalletResponse` is when we click the stake button and waiting for ...well... wallet response :)

So if we go with `loading` then we should rename this particular variable in different places also

totraev: > So if we go with loading then we should rename this particular variable in different places also

@gbarkhatov this is not true. Why our modal should know anything about `WalletResponse`? You can use loading state along side with other var names:

```
<Modal loading={awaitingWalletResponse} />
```

PR #340, opened 2024-11-13
Display phase 2 delegations

jrwbabylonlab:
image

100% Not the perfect PR for handling the delegations but our current simple-staking logic on delegations are way too messy for me to untangle. 
I would prefer get this out of the door so it unblock the rest of staking flow: stake, unbonding, withdraw etc.


gbarkhatov (approved):

totraev (approved):

src/app/api/getDelegationsV2.ts
```diff
@@ -51,19 +54,25 @@ export const getDelegationsV2 = async (
...
-      unbondingTx: apiDelegation.unbonding_tx,
-    }),
+    (apiDelegation: DelegationV2API): DelegationV2 => {
+      const state = DelegationV2StakingStateMap[apiDelegation.state];
+      if (!state) {
+        throw new Error(`Unknown delegation state: ${apiDelegation.state}`);
+      }
+      return {
+        finalityProviderBtcPksHex: apiDelegation.finality_provider_btc_pks_hex,
+        paramsVersion: parseInt(apiDelegation.params_version),
```
jeremy-babylonlabs, line 64: hmm should we return string directly from the api side so we don't need to parse on the FE?

jeremy-babylonlabs (approved):
lgtm with small nit comments

PR #364, opened 2024-11-18
fix: OKX signature tweak signer

gbarkhatov:
Fixes OKX `3.29.26` version `disableTweakSigner`

### DIFF
--- a/src/utils/wallet/providers/okx_wallet.ts
+++ b/src/utils/wallet/providers/okx_wallet.ts
@@ -112,7 +112,19 @@ export class OKXWallet extends WalletProvider {
       throw new Error("OKX Wallet not connected");
     }
     // Use signPsbt since it shows the fees
-    return await this.bitcoinNetworkProvider.signPsbt(psbtHex);
+    return await this.bitcoinNetworkProvider.signPsbt(
+      psbtHex,
+      // Required on 3.29.26 version
+      {
+        toSignInputs: [
+          {
+            index: 0,
+            address: this.okxWalletInfo.address,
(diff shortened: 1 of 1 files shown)


totraev (approved):

src/utils/wallet/providers/okx_wallet.ts
```diff
@@ -112,7 +112,19 @@ export class OKXWallet extends WalletProvider {
...
     }
     // Use signPsbt since it shows the fees
-    return await this.bitcoinNetworkProvider.signPsbt(psbtHex);
+    return await this.bitcoinNetworkProvider.signPsbt(
+      psbtHex,
+      // Required on 3.29.26 version
+      {
+        toSignInputs: [
+          {
+            index: 0,
```
vitsalis, line 121: This looks hacky. `signPsbt` is being used to not only sign transactions with a single input. Also, what if the input is not a taproot one?

vitsalis (changes requested):

src/utils/wallet/providers/okx_wallet.ts
gbarkhatov: Agree, but it's not failing to sign the psbt
We can do a list with options in `src/app/common/utils/psbt.ts`. This `signPsbtTransaction` is used only in Unbonding and Withdrawing

So if it's OKX and we are unbonding or withdrawing, we can pass the boolean flag

Based on that flag OKX will check whether the wallet is on Taproot, and if so - add the needed options

---

Another solution is for OKX to disable the changes, but question is what is more future-proof?

PR #369, opened 2024-11-20
feat: reduce signing in eoi

jrwbabylonlab:


totraev (approved):

gbarkhatov:
@jrwbabylonlab please ping me when it's ready for the review

src/app/components/Modals/FeedbackModal.tsx
```diff
@@ -142,7 +142,7 @@ export const FeedbackModal: React.FC<FeedbackModalProps> = ({
         </h3>
         <button
           className="btn btn-circle btn-ghost btn-sm"
-          onClick={() => onClose(false)}
```
gbarkhatov, line 145: I remember we had to use such flags, @jeremy-babylonlabs maybe you remember better
Is it okay to remove them now?

gbarkhatov (approved):
Looks good, but cannot be checked on the UI side

PR #374, opened 2024-11-23
Tong/refine UI 1

supertong:
A few things I would like to call out:
* This PR only covers header, fooder and background.
* Using the `bbn-core-ui/tailwind.config.js` as preset has caused some conflicts with the `daisyui` plugin. To avoid the rabbithole of replacing `daisyui`, I simply copied over the primary color from the shared config.
* I don't find some of the primary colors in dark theme is available in the shared tailwin config so I hard coded them for now. 

Demo:
image

image
image

image

image

image


public/footer-rectangle-dark.svg
totraev, line 1: lets avoid of using any bg images here, cause it's not very well solution for responsive design. We can implement this UI by using absolutely positioned `div` element or pseudo elements like `:before` & `:after`

src/app/components/Header/Header.tsx
```diff
@@ -16,8 +16,8 @@ export const Header = () => {
 
   return (
     <nav>
-      <div className="bg-base-300 shadow-sm">
-        <div className="container mx-auto flex w-full items-center justify-between gap-4 p-6 pb-4 md:pb-6">
+      <section className="bg-primary-main dark:bg-[#181818] h-[300px] mb-[-188px]">
```
totraev, line 19: @tong Don't use `dark:` modifier and don't try to implement dark mode atm. We gonna reuse the same color variable names by using different color palette. For better understanding check this lib https://www.npmjs.com/package/tw-colors

src/app/components/Header/Header.tsx
```diff
@@ -26,26 +26,29 @@ export const Header = () => {
...
+              loading={loading}
+              onConnect={open}
+              address={address}
+              btcWalletBalanceSat={totalBalance}
+              onDisconnect={disconnect}
+            />
+            <ThemeToggle />
+          </div>
+          <div
+            className={`${address && "justify-end p-6 pt-0"}container mx-auto flex w-full items-center gap-4 md:hidden md:p-0`}
```
totraev, line 40: use `twMerge` or `twJoin` instead

src/app/components/Header/SimplifiedHeader.tsx
```diff
@@ -6,8 +6,8 @@ import { TestingInfo } from "../TestingInfo/TestingInfo";
 export const SimplifiedHeader = () => {
   return (
     <nav>
-      <div className="bg-base-300 shadow-sm">
-        <div className="container mx-auto flex w-full items-center justify-between gap-4 p-6 pb-4 md:pb-6">
+      <section className="bg-primary-main dark:bg-[#181818] h-[300px] mb-[-188px]">
+        <div className="container h-[84px] py-[22px] px-6 mx-auto flex items-center justify-between">
```
totraev, line 10: lets better use `h-20` and `py-6` instead. they are not exact equal to this values but its better to avoid "custom" values

totraev (changes requested):

src/app/components/Header/Header.tsx
supertong: That's cool. I will remove the `dark:` modifier and ignore dark theme for now.

public/footer-rectangle-dark.svg
supertong: Can do. 👍

src/app/components/Header/Header.tsx
```diff
@@ -16,8 +18,8 @@ export const Header = () => {
 
   return (
     <nav>
-      <div className="bg-base-300 shadow-sm">
-        <div className="container mx-auto flex w-full items-center justify-between gap-4 p-6 pb-4 md:pb-6">
+      <section className="bg-primary-main h-[300px] mb-[-188px]">
```
totraev, line 21:
```suggestion
      <section className="bg-primary-main h-[300px] -mb-[188px]">
```

totraev (approved):

PR #378, opened 2024-11-24
Tong/use preset from core UI

supertong:
Managed to find a way to use preset and DaisyUI together without much conflict. Hopefully this would give us a smoother transition to our design system UI.


totraev (approved):

jeremy-babylonlabs (approved):

jeremy-babylonlabs:
lgtm~

PR #391, opened 2024-11-26
fix: update the states based on what backend return

jrwbabylonlab:


src/app/types/delegationsV2.ts
```diff
@@ -18,31 +18,50 @@ export interface DelegationV2 {
...
-  INTERMEDIATE_PENDING_CONFIRMATION:
-    DelegationV2StakingState.INTERMEDIATE_PENDING_CONFIRMATION,
-  INTERMEDIATE_UNBONDING: DelegationV2StakingState.INTERMEDIATE_UNBONDING,
-  INTERMEDIATE_WITHDRAWAL: DelegationV2StakingState.INTERMEDIATE_WITHDRAWAL,
+export const getDelegationV2StakingState = (
+  state: string,
+): DelegationV2StakingState => {
+  const validState = Object.values(DelegationV2StakingState).find(
+    (enumState) => enumState === state,
+  );
```
totraev, line 58:
```
export const getDelegationV2StakingState = (
  state: string,
): DelegationV2StakingState => {
  const validState = DelegationV2StakingState[state as DelegationV2StakingState];
 ```

PR #397, opened 2024-11-27
feat: use network info instead of params to make it more generic

jrwbabylonlab:
There is also a backend PR https://github.com/babylonlabs-io/staking-api-service/pull/148

This PR make the way for us to disable the staking form when staking is not yet allowed for public except those in the allowlist

### DIFF
--- a/src/app/types/networkInfo.ts
+++ b/src/app/types/networkInfo.ts
@@ -40,8 +40,3 @@ export interface NetworkInfo {
   stakingStatus: StakingStatus;
   params: Params;
 }
-
-// export interface NetworkInfoAPI {
-//   stakingStatus: StakingStatus;
-//   params: Params;
-// }


gbarkhatov (approved):

jeremy-babylonlabs (approved):

src/app/types/networkInfo.ts
```diff
@@ -27,7 +27,21 @@ export interface BtcEpochCheckParams {
...
   bbnStakingParams: BbnStakingParams;
   btcEpochCheckParams: BtcEpochCheckParams;
 }
+
+export interface NetworkInfo {
+  stakingStatus: StakingStatus;
+  params: Params;
+}
+
+// export interface NetworkInfoAPI {
```
gbarkhatov, line 44: Do we need it?

gbarkhatov (approved):

PR #406, opened 2024-11-28
feat: add eligibility for transition in delegations and update UI acc…

jeremy-babylonlabs:
- Introduced `isEligibleForTransition` property in the Delegation API and updated related components to reflect this new state
- Enhanced the Delegation component to conditionally render action buttons based on eligibility
- Updated the Delegations component to prevent rendering when there are no delegations
- Modified the Staking component to check if staking is open before rendering the staking form
- Added handling for the new `SLASHED` state in the delegation types


gbarkhatov (approved):

gbarkhatov:
@jeremy-babylonlabs build is failing

PR #422, opened 2024-12-02
feat: add tomo connector

totraev:


gbarkhatov (approved):
We can merge it since native wallets won't work without this PR, but if we use Tomo OKX and Tomo Keplr
https://i.imgur.com/L8dAXhp.png

@jrwbabylonlab I think you faced this issue before?

totraev:
> Question: Why not have the tomo inside the wallet connect component?

@jrwbabylonlab I faced several issues while was trying to install Tomo inside `bbn-wallet-connect`. I spent whole Saturday on it, so I decided to keep it inside `simple-staking` for now. but I agree that we need move it under `bbn-wallet-connect` after we resolve all our issues

totraev:
Issues with Tomo inside `bbn-wallet-connect`:

1. Buffer is not defined:
Screenshot 2024-12-01 at 09 18 30
Tomo team:

> The project requires buffer and process. If you're using Webpack, you can refer to this configuration: https://webpack.js.org/configuration/resolve/#resolvefallback.
> You can also check this code for a simple workaround: https://github.com/UnyxTech/tomo-wallet-provider/blob/main/src/demo/Inject.tsx.

2. Tomo's providers use hardcoded window object to access wallet providers
Screenshot 2024-12-03 at 08 46 06
This would not work inside `Storybook` cause it runs all components inside `iframe`

3. Tomo's modal does not use react portal what makes it really hard to use inside Storybook

PR #440, opened 2024-12-03
add btc usd price

jeremy-babylonlabs:


src/app/components/Stats/Stats.tsx
```diff
@@ -135,7 +137,14 @@ export const Stats: React.FC = () => {
       {
         title: "Confirmed TVL",
         value: stakingStats?.activeTVLSat
-          ? `${maxDecimals(satoshiToBtc(stakingStats.activeTVLSat), 2)} ${coinName}`
+          ? `${maxDecimals(satoshiToBtc(stakingStats.activeTVLSat), 2)} ${coinName}${
+              stakingStats.btcPriceUsd
```
vitsalis, line 141: On testnet this should always be `$0`

vitsalis: I don't mind not showing at all. What's important is that we do not show any real dollar values next to test tokens.

src/utils/formatAmount.ts
```diff
@@ -0,0 +1,32 @@
+export function formatAmount(amount: number): string {
```
totraev, line 1: Use build in [formatters](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat)

totraev (approved):

src/app/components/Stats/Stats.tsx
jeremy-babylonlabs: understood, I will merge this PR as is for now. In the public facing testnet we will not add the `btcPriceUsd` in the config from the api side cc:@filippos47

PR #470, opened 2024-12-07
Tong/436 feedback modal refine

supertong:
It looks like the CongraModal should be part of FeedbackModal so I have remove that one and move it into FeedbackModal.
I do not find the design for FeedbackModal when the staking is cancelled, so I made my best effort to make it look nice.

image

image


jeremy-babylonlabs (approved):

jeremy-babylonlabs:
design is neat, nice!

PR #475, opened 2024-12-09
feat: add phase-1 unbonding method

jrwbabylonlab:
This PR only creates the necessary hook method for unbonding phase-1 delegation.

@jeremy-babylonlabs will work on putting together with the FE UI modal.

### DIFF
--- a/src/app/components/Modals/UnbondModal.tsx
+++ b/src/app/components/Modals/UnbondModal.tsx
@@ -18,6 +18,7 @@ interface UnbondModalProps {
   onProceed: () => void;
   awaitingWalletResponse: boolean;
 }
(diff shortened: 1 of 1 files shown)


src/app/components/Modals/UnbondModal.tsx
```diff
@@ -24,6 +25,7 @@ export const UnbondModal = ({
   onProceed,
   awaitingWalletResponse,
 }: UnbondModalProps) => {
+  const { networkName, coinName } = getNetworkConfig();
```
totraev, line 28: No need to call it inside render function

totraev (approved):

PR #488, opened 2024-12-10
feat: disable stake button if utxo not available

jrwbabylonlab:
image

### DIFF
--- a/src/app/state/index.tsx
+++ b/src/app/state/index.tsx
@@ -35,8 +35,8 @@ const { StateProvider, useState: useApplicationState } =
     isError: false,
     totalBalance: 0,
     ordinalsExcluded: true,
-    includeOrdinals: () => { },
-    excludeOrdinals: () => { },
+    includeOrdinals: () => {},
+    excludeOrdinals: () => {},
   });
 
 export function AppState({ children }: PropsWithChildren) {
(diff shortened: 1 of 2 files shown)


totraev (changes requested):
Will update later today

gbarkhatov (approved):

src/app/state/index.tsx
```diff
@@ -21,6 +21,7 @@ const STATE_LIST = [DelegationState, DelegationV2State];
 
 export interface AppState {
   availableUTXOs?: UTXO[];
+  allUTXOs?: UTXO[];
```
totraev, line 24: @jrwbabylonlab it's not clear to me why do you use `allUTXOs` instead of `availableUTXOs`?

totraev (changes requested):

totraev:
> What need to be updated?

Don't write any domain specific logic inside render function directly. Use my fixes as a reference for future PRs

totraev:
@jrwbabylonlab can you pls check my changes? I didn't test them

totraev:
@jrwbabylonlab I see another problem: everytime we submit a staking tx we need invalidate utxos' cache and refetch them

src/app/state/index.tsx
totraev: @jrwbabylonlab This is expected behaviour IMO. If user decided to filter out ordinals (to save them) we should not let him to stake his previously created delegation with ordinals

totraev (approved):

totraev:
@jrwbabylonlab approved, but pls test before merge

PR #493, opened 2024-12-11
fix search not applied on production environment

jeremy-babylonlabs:
this pr fixes the issue where `/?fp=xxx` filter is not properly applied on the list

### DIFF
--- a/src/app/components/Staking/FinalityProviders/FinalityProviderSearch.tsx
+++ b/src/app/components/Staking/FinalityProviders/FinalityProviderSearch.tsx
@@ -14,11 +14,19 @@ export const FinalityProviderSearch: React.FC<FinalityProviderSearchProps> = ({
   const searchParams = useSearchParams();
   const initialSearchFp = searchParams.get("fp");
 
-  const [searchTerm, setSearchTerm] = useState(initialSearchFp);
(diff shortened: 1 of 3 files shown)


src/app/components/Staking/FinalityProviders/FinalityProviderSearch.tsx
```diff
@@ -14,11 +14,19 @@ export const FinalityProviderSearch: React.FC<FinalityProviderSearchProps> = ({
...
-    onSearch(debouncedSearchTerm ?? "");
+    if (initialSearchFp) {
+      onSearch(initialSearchFp);
+    }
+  }, [initialSearchFp, onSearch]);
+
+  // Effect for handling subsequent searches
+  useEffect(() => {
+    onSearch(debouncedSearchTerm);
   }, [debouncedSearchTerm, onSearch]);
```
totraev, line 30: What issue we're trying to fix? Previous implementation works for me on my local

totraev (approved):

src/app/components/Staking/FinalityProviders/FinalityProviderSearch.tsx
jeremy-babylonlabs: It doesn't work on production environment

PR #498, opened 2024-12-12
fix: inception unable to handle  months and years

jrwbabylonlab:
The issue was that we weren’t handling months and years properly, but now we do. Additionally, I’ve removed hours and minutes when displaying months or years, as including too much detail would make the field unnecessarily lengthy.

image

image

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "simple-staking",
-  "version": "0.3.20",
+  "version": "0.3.21",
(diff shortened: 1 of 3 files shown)


gbarkhatov (approved):

jeremy-babylonlabs (approved):
lgtm!

totraev:
Question: why do we need this `durationTillNow` method if `data-fns` already has it? https://date-fns.org/v4.1.0/docs/formatDistanceToNow

PR #505, opened 2024-12-15
Tong/476 error modal

supertong:
image

Didn't find the design so I only replaced the component with `bbn-core-ui`.


gbarkhatov (approved):
According to the screenshot, I would make the header smaller https://i.imgur.com/3PjUlex.png

PR #514, opened 2024-12-17
fix: supporting sending query param in repeat format

jrwbabylonlab:


src/app/api/apiWrapper.ts
```diff
@@ -1,45 +1,44 @@
...
-          },
-      {
-        timeout: timeout || 0, // 0 is no timeout
-      },
-    );
+    if (method === "GET") {
+      return await axios.get(url, axiosConfig);
+    } else {
+      return await axios.post(url, params?.body, axiosConfig);
+    }
```
totraev, line 37: you can use `axios.request` instead

totraev (approved):

jeremy-babylonlabs (approved):

PR #533, opened 2024-12-20
fix: use param data instead of hardcoded unbonding fee and confirmati…

jrwbabylonlab:
This PR fix the issue where we using mocked data for unbonding fee and confirmation depth

### DIFF
--- a/src/app/components/Modals/UnbondModal.tsx
+++ b/src/app/components/Modals/UnbondModal.tsx
@@ -23,12 +23,14 @@ export const UnbondModal = ({
   processing,
 }: UnbondModalProps) => {
   const { data: networkInfo } = useNetworkInfo();
+  if (!networkInfo) throw new Error("Network info not found");
+
   const unbondingTime = blocksToDisplayTime(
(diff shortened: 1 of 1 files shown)


src/app/components/Modals/UnbondModal.tsx
```diff
@@ -18,6 +22,17 @@ export const UnbondModal = ({
...
+  const { data: networkInfo } = useNetworkInfo();
+  const unbondingTime = blocksToDisplayTime(
+    networkInfo?.params.bbnStakingParams?.latestParam?.unbondingTime,
+  );
+  const unbondingFeeBtc = maxDecimals(
+    satoshiToBtc(
+      networkInfo?.params.bbnStakingParams?.latestParam?.unbondingFeeSat ?? 0,
+    ),
+    8,
+  );
```
totraev, line 34: This is not right place for this logic. Pass it via component's props, encapsulate into related service

totraev (approved):

jeremy-babylonlabs (approved):

PR #542, opened 2024-12-22
Tong/458 staking not available

supertong:
image

I have refactored the `Message` component so that it can be reused. Let me know what do you think.


src/app/components/Staking/Form/States/Message.tsx
```diff
@@ -1,27 +1,30 @@
+import { Heading, Text } from "@babylonlabs-io/bbn-core-ui";
 import Image from "next/image";
 
 interface MessageProps {
   title: string;
-  messages: string[];
```
supertong, line 5: Update this to just a string but we can still use the string literal template to do line breaks if we want to.

totraev (approved):

PR #543, opened 2024-12-22
add claim reward modal

supertong:
image

image

When the "Claim" button is clicked, the modal in the above image shoud display.
Note: I have place in a few placeholders since I have no idea where to get those value from.

When the "Proceed" button is clicked on the modal, it will send a notification and when the claim action was successful, a success notification will be send after that.
Note: We might need a redesign of the notification banners but I have no idea if the design is ready so I will leave it for now.

### DIFF
--- a/src/app/components/Modals/ClaimRewardModal.tsx
+++ b/src/app/components/Modals/ClaimRewardModal.tsx
@@ -1,0 +1,58 @@
+import { Heading, Text } from "@babylonlabs-io/bbn-core-ui";
+import { PropsWithChildren } from "react";
+
+import { ConfirmationModal } from "./ConfirmationModal";
+
+interface ConfirmationModalProps {
+  open: boolean;
+  onClose: () => void;
+  onSubmit: () => void;
+  receivingValue: string;
+  address: string;
+  transactionFee: string;
+}
+
+export const ClaimRewardModal = ({
+  open,
+  onClose,
+  onSubmit,
+  receivingValue,
+  address,
+  transactionFee,
+}: PropsWithChildren<ConfirmationModalProps>) => {
+  return (
+    <ConfirmationModal
+      open={open}
+      processing={false}
+      title="Claim tBABY Reward"
+      onClose={onClose}
+      onSubmit={onSubmit}
+    >
+      <div className="flex flex-col mt-8 gap-10">
+        <div className="flex flex-col gap-4 divide-y divide-inherit">
+          <div className="flex flex-row items-center justify-between">
+            <Text variant="body1" className="text-center">
+              Receiving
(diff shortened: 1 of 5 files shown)


src/app/components/Modals/ClaimRewardModal.tsx
```diff
@@ -0,0 +1,74 @@
...
+}: PropsWithChildren<ConfirmationModalProps>) => {
+  const [transactionFee, setTransactionFee] = useState<number>(0);
+
+  useEffect(() => {
+    const fetchTransactionFee = async () => {
+      const fee = await getTransactionFee();
+      setTransactionFee(fee);
+    };
+    fetchTransactionFee();
+  }, [getTransactionFee]);
```
totraev, line 33: Don't write domain logic inside render function directly. Encapsulate it into related service

src/app/components/Notification/NotificationText.tsx
```diff
@@ -4,8 +4,6 @@ interface Props {
 
 export const NotificationText = ({ children }: Props) => {
   return (
-    <span className="text-sm font-normal dark:text-[#AFAFAF] text-[#303030]">
-      {children}
-    </span>
+    <span className="text-sm font-normal text-primary-dark">{children}</span>
```
totraev, line 7: `<Text variant="body2" classname="text-primary-dark">{children}<Text/>`

src/app/components/Notification/NotificationTitle.tsx
```diff
@@ -4,8 +4,6 @@ interface Props {
 
 export const NotificationTitle = ({ children }: Props) => {
   return (
-    <span className="text-base font-bold dark:text-white text-black">
-      {children}
-    </span>
+    <span className="text-base font-bold text-primary-dark">{children}</span>
```
totraev, line 7: `<Text variant="body1" classname="text-primary-dark">{children}<Text/>`

src/app/components/PersonalBalance/PersonalBalance.tsx
```diff
@@ -42,6 +53,8 @@ export function PersonalBalance() {
     return null;
   }
 
+  const rewardBalance = ubbnToBbn(rewardsQuery.data ?? 0);
```
totraev, line 56: Domain logic inside render function

totraev (changes requested):

src/app/components/Modals/ClaimRewardModal.tsx
totraev: @jrwbabylonlab let's discuss this on our call today as well

jeremy-babylonlabs (approved):

PR #568, opened 2024-12-30
add transition register to bbn chain modal

### LINKED ISSUE #561: [UI] Transition Register to BBN Chain Modal
Add modal and trigger signing

image

jeremy-babylonlabs:
Closes #561 

The signing modal should share the implementation that is refactored in the #567. So I will wait for the PR to merge and update this PR.


src/utils/trim.ts
totraev, line 1: nit: this function looks confusing. Usually `trim` means remove empty lines before and after the string (`truncate` is more suitable name). Also it shouldn't return values like '-' or just '...' cause such behaviour is not responsibility of this function

totraev (approved):

src/app/components/Modals/RegistrationModal/RegistrationStartModal.tsx
totraev, line 1: Use `SubmitModal` instead. Check the implementation of `PreviewModal` and `StakeModal`

src/app/components/Modals/RegistrationModal/RegistrationEndModal.tsx
totraev, line 1: Extend `SubmitModal`

src/app/hooks/services/useRegistrationService.ts
```diff
@@ -0,0 +1,71 @@
...
+              case SigningStep.PROOF_OF_POSSESSION:
+                setStep("proof-of-possession");
+                break;
+              case SigningStep.SIGN_BBN:
+                setStep("sign-bbn");
+                break;
+              case SigningStep.SEND_BBN:
+                setStep("send-bbn");
+                break;
+            }
```
totraev, line 51: Don't use `switch/case`. It's a bad practice

src/app/state/RegistrationState.tsx
totraev, line 1: It should be a part of `DelegationState.tsx`

src/app/hooks/services/useRegistrationService.ts
totraev, line 51:
```suggestion
            setStep(step)
```

src/app/state/RegistrationState.tsx
```diff
@@ -0,0 +1,80 @@
...
+export interface RegistrationFormFields {
+  stakingTxHex: string;
+  startHeight: number;
+  stakingInput: BtcStakingInputs;
+}
+
+export interface RegistrationState {
+  processing: boolean;
+  step?: RegistrationStep;
+  formData?: RegistrationFormFields;
```
totraev, line 25: We don't have a form for delegations, you can save whole delegation here or just a delegation id

totraev (approved):

src/app/state/DelegationState.tsx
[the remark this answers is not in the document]
totraev: No, we have both DelegationState and DelegationV2State

PR #574, opened 2025-01-03
feat: show correct total balance

jrwbabylonlab:
image

This is a temporary solution while @gusin13 working on a proper API based solution
We will make the API to return large num of delegations ~100 to workaround this.


jeremy-babylonlabs (approved):

src/app/components/PersonalBalance/PersonalBalance.tsx
```diff
@@ -61,7 +62,7 @@ export function PersonalBalance() {
   const rewardBalance = ubbnToBbn(rewardsQuery.data ?? 0);
 
   const stakerStakedBalance = getStakedBalance();
-  const totalBalance = (btcBalance ?? 0) + stakerStakedBalance;
+  const totalBalance = calculateTotalBalance(btcBalance, stakerStakedBalance);
```
totraev, line 65: useMemo

src/utils/balance.ts
```diff
@@ -0,0 +1,51 @@
...
+ */
+export function calculateTotalBalance(
+  btcBalance: number | undefined,
+  stakedBalance: number,
+): number {
+  try {
+    const validatedBtcBalance = validateBalance(btcBalance);
+    const validatedStakedBalance = validateBalance(stakedBalance);
+
+    // Use BigInt for safe integer arithmetic
```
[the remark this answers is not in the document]
jeremy-babylonlabs: I worry there might be precision issues or even overflow issues when we dont use bigint. I think its common practice when we do arithmetic on these large numbers to keep things precise.

We might need to check if we are doing arithmetics on other places too to keep things consistent.

src/utils/balance.ts
```diff
@@ -0,0 +1,51 @@
...
+ * @param stakedBalance - Staked balance amount
+ * @returns Combined total balance
+ */
+export function calculateTotalBalance(
+  btcBalance: number | undefined,
+  stakedBalance: number,
+): number {
+  try {
+    const validatedBtcBalance = validateBalance(btcBalance);
+    const validatedStakedBalance = validateBalance(stakedBalance);
```
totraev, line 35: Remove validations, data should be pre-validated

src/utils/balance.ts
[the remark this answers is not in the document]
totraev: BigInt is not the best choice, we need use something like this  https://mikemcl.github.io/bignumber.js/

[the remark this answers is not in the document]
totraev: BigInt still has limits

src/utils/balance.ts
```diff
@@ -0,0 +1,51 @@
...
+
+    // Use BigInt for safe integer arithmetic
+    const total =
+      BigInt(validatedBtcBalance.amount) +
+      BigInt(validatedStakedBalance.amount);
+
+    // Check for overflow
+    if (total > BigInt(Number.MAX_SAFE_INTEGER)) {
+      throw new Error("Balance error: Balance overflow detected");
+    }
```
totraev, line 45: don't throw validation error inside render function

src/utils/balance.ts
```diff
@@ -0,0 +1,51 @@
...
+
+    return Number(total);
+  } catch (error) {
+    throw new Error(`Balance error: ${error}`);
```
totraev, line 49: the same as above

jeremy-babylonlabs:
lgtm

totraev (approved):

PR #611, opened 2025-01-09
improve fp search to allow search to not scope to showing tab

jeremy-babylonlabs:
https://github.com/user-attachments/assets/1eeb1a67-467a-4a2a-b027-f53497796374

### DIFF
--- a/src/app/components/Staking/FinalityProviders/FinalityProviderTable.tsx
+++ b/src/app/components/Staking/FinalityProviders/FinalityProviderTable.tsx
@@ -82,7 +82,6 @@ export const FinalityProviderTable = ({
   return (
     <div className="h-[21rem] overflow-y-auto ">
       <Table
-        key={`${filter.search}-${filter.status}`}
         data={tableData}
         columns={finalityProviderColumns}
         loading={isFetching}
--- a/src/app/components/Staking/FinalityProviders/FinalityProviderTable.tsx
+++ b/src/app/components/Staking/FinalityProviders/FinalityProviderTable.tsx
@@ -20,10 +20,9 @@ export const FinalityProviderTable = ({
     isFetching,
     finalityProviders,
     hasNextPage,
-    fetchNextPage,
-    searchValue,
-    filterValue,
     hasError,
+    filter,
+    fetchNextPage,
     isRowSelectable,
   } = useFinalityProviderState();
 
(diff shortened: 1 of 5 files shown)


totraev (changes requested):
Hold

src/app/state/FinalityProviderState.tsx
totraev, line 1: It's not clear to me why do you need `isInitialLoad` + `previousFilterValue`

src/app/state/FinalityProviderState.tsx
```diff
@@ -141,7 +150,14 @@ function FinalityProviderStateInner({ children }: PropsWithChildren) {
...
-      return true;
+      const isActive = fp.state === FinalityProviderStateEnum.ACTIVE;
+      const isInactive = inactiveStatuses.has(fp.state);
+
+      return filterValue === "active"
+        ? isActive
+        : filterValue === "inactive"
+          ? isInactive
+          : true;
     });
```
totraev, line 161: Double termary operator

totraev (changes requested):

src/app/state/FinalityProviderState.tsx
jeremy-babylonlabs: with `previousFilterValue`, clearing a search would return to the previous filter state
with isInitialLoad, it act as a flag to handle the special case when the page is loaded with a URL param (?fp=xxx)

totraev: @jeremy-babylonlabs you don't need them, you can replace all `flags` and `if/else` statements by a single strategy pattern, pls check my commit. And pls make some additional tests on my changes

src/app/components/Staking/FinalityProviders/FinalityProviderTable.tsx
```diff
@@ -83,7 +82,7 @@ export const FinalityProviderTable = ({
   return (
     <div className="h-[21rem] overflow-y-auto ">
       <Table
-        key={`${searchValue}-${filterValue}`}
+        key={`${filter.search}-${filter.status}`}
```
totraev, line 85: I'm not fully following why you need this. @jeremy-babylonlabs can you pls explain?

totraev: atm on every state change we fully remount (!) our table here

totraev (approved):

PR #613, opened 2025-01-10
Fix slashing tooltip

jeremy-babylonlabs:
Show Invalid status and corresponding tooltip when EOI delegations are slashed

### DIFF
--- a/src/components/delegations/DelegationList/components/Status.tsx
+++ b/src/components/delegations/DelegationList/components/Status.tsx
@@ -25,12 +25,15 @@ interface StatusTooltipProps {
   amount?: number;
   coinName?: string;
   slashingAmount?: number;
+  startHeight?: number;
 }
+
+const { coinName } = getNetworkConfigBTC();
 
 const STATUS_LABELS = {
   PENDING: "Pending",
   VERIFIED: "Verified",
-  PENDING_BTC_CONFIRMATION: "Pending BTC Confirmation",
+  PENDING_BTC_CONFIRMATION: `Pending ${coinName} Confirmation`,
   ACTIVE: "Active",
   UNBONDING: "Unbonding",
   WITHDRAWABLE: "Withdrawable",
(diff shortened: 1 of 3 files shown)


totraev (changes requested):
Hold this one

src/components/delegations/DelegationList/components/Status.tsx
```diff
@@ -183,6 +228,8 @@ export function Status({ delegation }: StatusProps) {
     params: networkInfo?.params,
     amount: delegation.stakingAmount,
     coinName,
+    slashingAmount,
+    startHeight: delegation.startHeight,
```
totraev, line 232: Instead of passing every value individually just use entities/interfaces:
`GlobalParams`, `Delegation`, `Config`

src/utils/delegations/slashing.ts
```diff
@@ -0,0 +1,22 @@
...
+ * @param param - The staking param.
+ * @returns The slashing amount.
+ */
+export const getSlashingAmount = (
+  stakingAmount: number,
+  param: BbnStakingParamsVersion,
+) => {
+  if (!param.slashing) {
+    throw new Error("Slashing param not found");
+  }
```
totraev, line 16: don't throw errors in render function

src/components/delegations/DelegationList/components/Status.tsx
```diff
@@ -175,6 +213,13 @@ const STATUSES: Record<
 export function Status({ delegation }: StatusProps) {
   const { networkInfo } = useAppState();
   const { coinName } = getNetworkConfigBTC();
+
+  const param = getBbnParamByVersion(
+    delegation.paramsVersion,
+    networkInfo?.params.bbnStakingParams.versions || [],
+  );
```
totraev, line 220: To avoid performance issues we need move it from `Status` component

totraev (changes requested):

totraev (approved):
Approved with some minor changes. @jeremy-babylonlabs can you double check that I didn't break anything

PR #619, opened 2025-01-10
chore: use gas price from env

jrwbabylonlab:


src/config/index.ts
```diff
@@ -24,3 +24,11 @@ export const getBtcNetwork = (): Network => {
 
 export const IS_FIXED_TERM_FIELD =
   process.env.NEXT_PUBLIC_FIXED_STAKING_TERM === "true";
+
+export const BBN_GAS_PRICE = (() => {
+  const price = Number(process.env.NEXT_PUBLIC_BBN_GAS_PRICE) || 0.002;
```
totraev, line 29: use `parseFloat` instead

src/config/index.ts
```diff
@@ -24,3 +24,11 @@ export const getBtcNetwork = (): Network => {
...
+  if (isNaN(price) || price <= 0 || price >= 1) {
+    return 0.002; // fallback to default if invalid
```
totraev, line 31: It would be better to use yet another constant instead of magic number + comment (Like DEFAULT_BBN_GAS_PRICE)

totraev (dismissed):

totraev (approved):

PR #624, opened 2025-01-12
display phase 1 fp name

### LINKED ISSUE #519: Show FP name for phase-1 delegation
Feedback from user is that they would like to know which FP their delegation was selected.

We already have this piece of data in FE. so might as well display it

jeremy-babylonlabs:
closes #519 

this PR changes the pending registration tooltip and show phase 1 fp name


totraev:
Sorry, this PR will not pass

PR #627, opened 2025-01-13
fix: should not show staking form if not enabled

jrwbabylonlab:
image


totraev:
It's better to do that through state, check my PR

totraev:
@jrwbabylonlab can you pls recheck whole flow. I didn't test my changes

totraev (approved):

PR #632, opened 2025-01-14
fix: should show all UTXOs for total balance

jrwbabylonlab:


src/app/state/index.tsx
```diff
@@ -53,10 +55,12 @@ const { StateProvider, useState: useApplicationState } =
 export function AppState({ children }: PropsWithChildren) {
   const { lockInscriptions: ordinalsExcluded, toggleLockInscriptions } =
     useInscriptionProvider();
+  const { getStakedBalance } = useDelegationV2State();
```
totraev, line 58: Don't use state inside state. The only exception is AppState, cause it kinda `global` state

totraev (approved):

PR #633, opened 2025-01-14
phase 1 fp name

jeremy-babylonlabs:
closes https://github.com/babylonlabs-io/simple-staking/issues/519

this PR changes the pending registration tooltip and show phase 1 fp name


src/app/state/FinalityProviderState.tsx
```diff
@@ -61,12 +70,19 @@ const defaultState: FinalityProviderState = {
   handleFilter: () => {},
   getFinalityProvider: () => null,
   fetchNextPage: () => {},
+  // V1
+  finalityProvidersV1: [],
+  hasNextPageV1: false,
+  isFetchingV1: false,
+  hasErrorV1: false,
+  getFinalityProviderV1: () => null,
+  fetchNextPageV1: () => {},
```
totraev, line 79: You don't need export all these props except `getFinalityProviderV1`

src/app/state/FinalityProviderState.tsx
```diff
@@ -61,12 +70,19 @@ const defaultState: FinalityProviderState = {
...
 };
 
 const { StateProvider, useState: useFpState } =
   createStateUtils<FinalityProviderState>(defaultState);
 
-function FinalityProviderStateInner({ children }: PropsWithChildren) {
+const FinalityProviderStateInner = ({ children }: PropsWithChildren) => {
```
totraev, line 85: Can you also explain why do you use Suspense here? We don't need it here, it's only used for loading

src/app/state/FinalityProviderState.tsx
```diff
@@ -147,10 +175,20 @@ function FinalityProviderStateInner({ children }: PropsWithChildren) {
...
     [data?.finalityProviders],
   );
 
+  const getFinalityProviderV1 = useCallback(
+    (btcPkHex: string) =>
+      dataV1?.finalityProviders.find(
+        (fp: FinalityProviderV1) => fp.btcPk === btcPkHex,
+      ) || null,
+    [dataV1?.finalityProviders],
+  );
```
totraev, line 190: These 2 methods are pretty slow especially if you use them inside render function. On every state change (let say on filtering) this method goes through FP list for every FP or Delegation in the table. You can replace fp list with fp map to optimise this (O(n) -> 0(1))

totraev (changes requested):

totraev:
Pls merge this PR first https://github.com/babylonlabs-io/simple-staking/pull/611 and than we can try resolve conflicts here

totraev (approved):
Good job!

PR #649, opened 2025-01-23
improve inscription safety messaging for better clarity

fuxingloh:
> downstream team testing and building on top of Babylon.

Anyway, this messaging could use better clarity.

```diff
- {ordinalsExcluded ? "Not using Inscriptions" : "Using Inscriptions"}
+ {ordinalsExcluded ? "Inscription Safety: Enabled" : "Inscription Safety: Disabled"}
```

It caught me off guard, and I immediately thought, "Not using Inscriptions" Protection?
But really, it means "Not using (my) Inscriptions". I think a better alternative is "Inscription Safety: Enabled" since you can't guarantee but can enable safety.


totraev:
I'll close PR for now

PR #672, opened 2025-01-31
fix rsc header vulnerability

### LINKED ISSUE #653: RS header issue
https://babylonlabsworkspace.slack.com/archives/C07ES6DG71P/p1737618004740089

jeremy-babylonlabs:
closes #653

### DIFF
--- a/src/middleware.ts
+++ b/src/middleware.ts
@@ -1,0 +1,31 @@
+import type { NextRequest } from "next/server";
+import { NextResponse } from "next/server";
+
+/**
+ * Headers that could enable cache poisoning attacks if exposed
+ */
+const UNSAFE_HEADERS = new Set([
+  "rsc", // Prevents RSC payload poisoning
+  "x-middleware-prefetch", // Prevents prefetch cache poisoning
+  "x-invoke-status", // Prevents status code override attacks
+  "next-router-state-tree", // Additional Next.js internal headers
+  "next-router-prefetch",
+]);
+
+export function middleware(request: NextRequest) {
+  const response = NextResponse.next();
+
+  // Remove potentially dangerous headers
+  Array.from(response.headers.keys()).forEach((key) => {
+    if (UNSAFE_HEADERS.has(key.toLowerCase())) {
+      response.headers.delete(key);
+    }
+  });
+
+  return response;
+}
+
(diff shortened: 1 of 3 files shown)


src/middleware.ts
```diff
@@ -0,0 +1,27 @@
+import type { NextRequest } from "next/server";
+import { NextResponse } from "next/server";
+
+// Security headers middleware
+export function sanitizeHeaders(headers: Headers): Headers {
+  const sanitized = new Headers(headers);
+  sanitized.delete("Rsc");
+  sanitized.delete("Next-Router-Prefetch");
+  sanitized.delete("Next-Router-State-Tree");
```
totraev, line 9: question: why do we need remove this?

jeremy-babylonlabs: these 2 are also internal headers for nextjs and exposes risk of manipulation

src/middleware.ts
```diff
@@ -0,0 +1,54 @@
...
+]);
+
+export function middleware(_request: NextRequest) {
+  const response = NextResponse.next();
+  Array.from(response.headers.keys()).forEach((key) => {
+    // Remove potentially dangerous headers and set security headers
+    if (UNSAFE_HEADERS.has(key.toLowerCase())) {
+      response.headers.delete(key);
+    }
+  });
```
totraev, line 39: I don't understand how the removal of `unsafe_headers` from the response (!) object can protect us?

src/middleware.ts
```diff
@@ -0,0 +1,54 @@
...
+
+  Object.entries(SECURITY_HEADERS).forEach(([key, value]) => {
+    response.headers.set(key, value);
+  });
```
totraev, line 43: If we completely disable CDN cache on this step then we create potential performance issues + create possibility for DDOS on our SSR server

src/middleware.ts
```diff
@@ -0,0 +1,54 @@
...
+
+  const SAFE_VARY = ["Accept-Encoding", "RSC", "x-middleware-prefetch"];
+  response.headers.set("Vary", SAFE_VARY.join(", "));
```
totraev, line 46: Cloudfare (do we use Cloudfare, right?) ignores `Vary` header

totraev (changes requested):

PR #680, opened 2025-02-07
Adopt sdk

jrwbabylonlab:


src/app/hooks/services/useRegistrationService.ts
```diff
@@ -65,14 +70,33 @@ export function useRegistrationService() {
         },
       };
 
-      await transitionPhase1Delegation(
+      const unsubscribe = subscribeToSigningSteps((step: SigningType) => {
```
totraev, line 73: don't subscribe to ee inside async/await function. Use useEffect instead:

```
useEffect(() => {
    const unsubscribe = subscribeToSigningSteps(\\...);
    
    return unsubscribe
}, [subscribeToSigningSteps]);
```

src/app/hooks/services/useRegistrationService.ts
```diff
@@ -65,14 +70,33 @@ export function useRegistrationService() {
...
+      const { signedBabylonTx } = await transitionPhase1Delegation(
         registrationData.stakingTxHex,
         registrationData.startHeight,
         registrationData.stakingInput,
-        async (step: SigningStep) => {
-          setStep(`registration-${step}`);
-        },
       );
+      // Send the transaction
+      setStep("registration-send-bbn");
```
totraev, line 96: this logic needs to be moved under ee

src/app/hooks/services/useStakingManagerService.ts
```diff
@@ -0,0 +1,67 @@
...
+  const {
+    network: btcNetwork,
+    connected: btcConnected,
+    signPsbt,
+    signMessage,
+  } = useBTCWallet();
+
+  const versionedParams = networkInfo?.params.bbnStakingParams?.versions;
+
+  const btcStakingManager = useMemo(() => {
```
totraev, line 23: This is super bad idea to create some class inside `useMemo`. You have race conditions here: your `useMemo` hook depends on `versionedParams` which can be updated at any time (thx to `react-query`). In this case your object will be recreated and it will lose its own state.

src/app/hooks/services/useStakingService.ts
```diff
@@ -62,31 +74,50 @@ export function useStakingService() {
...
+              goToStep(StakingStep.EOI_UNBONDING_SLASHING);
+              break;
+            case SigningType.PROOF_OF_POSSESSION:
+              goToStep(StakingStep.EOI_PROOF_OF_POSSESSION);
+              break;
+            case SigningType.CREATE_BTC_DELEGATION_MSG:
+              goToStep(StakingStep.EOI_SIGN_BBN);
+              break;
+          }
+        });
```
totraev, line 92: the same as above

src/app/state/StakingState.tsx
```diff
@@ -266,22 +266,21 @@ export function StakingState({ children }: PropsWithChildren) {
...
       }
 
-      if (stepName === "feedback-cancel") {
+      if (stepName === StakingStep.FEEDBACK_CANCEL) {
         if (cancelModalShown) {
           return;
         } else {
           setCancelModalShown(true);
         }
       }
```
totraev, line 283: too many `if/else` conditions

totraev (changes requested):

totraev (approved):

Issue #571, opened 2025-01-02
Warning modal when address + PK mismatch


jrwbabylonlab:
This ticket is for both phase-1 and 2

Phase-2 design https://www.figma.com/design/05UW9XBXoFWcKEEIB1ZetV/Design-File?node-id=5271-831296&t=pSGQw7xsdiqDpgkZ-0

PR #684, opened 2025-02-10
feat: validate address & pk

0xDazzer:
Closes: https://github.com/babylonlabs-io/simple-staking/issues/571


src/utils/wallet/index.ts
```diff
@@ -32,3 +34,56 @@ export const getPublicKeyNoCoord = (pkHex: string): Buffer => {
   const publicKey = Buffer.from(pkHex, "hex");
   return publicKey.subarray(1, 33);
 };
+
+export const getTaprootAddress = (publicKey: string, network: Network) => {
```
jeremy-babylonlabs, line 38: nit q, do we need to export this func?

0xDazzer: hmmm... good question. It might be needed for testing purpose

PR #689, opened 2025-02-15
add missing sentry config to docker image

supertong:
This might be the reason why it didn't fire anything in https://btcstaking.phase-2-devnet.babylonlabs.io/


jeremy-babylonlabs (approved):
Nice catch!

supertong:
nope, this solves dev-net not sending the event. We still need a new service to work around ad-blocker.

PR #716, opened 2025-02-28
chore: remove regex from the search filter

jrwbabylonlab:


0xDazzer (approved):
@jrwbabylonlab approved, but I'm not following why we need remove regex. Is it because of security audit?

PR #727, opened 2025-03-05
fix: persistent connection for Tomo wallets

0xDazzer:


src/app/context/wallet/CosmosWalletProvider.tsx
```diff
@@ -94,7 +99,7 @@ export const CosmosWalletProvider = ({ children }: PropsWithChildren) => {
         });
       }
     },
-    [handleError, rpc],
+    [handleError],
```
jonybur, line 102: Don't we need to keep `rpc` in the dependency list given that it is still used on line 82?

jonybur (changes requested):
Please check comment

jonybur (approved):

Issue #714, opened 2025-02-27
Show unconfirmed UTXO


jrwbabylonlab:
https://www.figma.com/design/05UW9XBXoFWcKEEIB1ZetV/Design-File?node-id=6022-358950&t=4rqsRL0TDqZjvSNZ-0#1146153688

PR #729, opened 2025-03-06
feat: adds unconfirmed UTXOs to balance

jonybur:
Shows unconfirmed UTXOs balance in the Stakable Balance panel

Screenshot 2025-03-06 at 16 35 52

https://github.com/babylonlabs-io/simple-staking/issues/714

### DIFF
--- a/src/app/components/Stats/StatItem.tsx
+++ b/src/app/components/Stats/StatItem.tsx
@@ -13,15 +13,20 @@ export enum LoadingStyle {
   ShowSpinner = "show-spinner",
   ShowSpinnerAndValue = "show-spinner-and-value",
 }
-const SPINNER_RENDERERS: Record<LoadingStyle, (value: string) => JSX.Element> = {
+
+const SPINNER_RENDERERS: Record<
+  LoadingStyle,
+  (value: string | JSX.Element) => JSX.Element
+> = {
   [LoadingStyle.ShowSpinner]: () => <Loader size={20} />,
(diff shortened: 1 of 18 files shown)


src/app/components/Stats/StatItem.tsx
```diff
@@ -38,8 +44,10 @@ export const StatItem = ({
     <ListItem
       {...props}
       title={title}
-      value={loading ? <Loader size={20} /> : value}
+      value={value}
       suffix={suffixEl}
+      loading={loading}
+      loadingStyle={loadingStyle}
     />
```
0xDazzer, line 51: I see that PR is still in progress but I wanna highlight one thing. You don't need modify `<ListItem />` from `core-ui`, cause current functionality was intended. Not every list item really need loading state (thats why it was not added to `core-ui`), but if you do need it somewhere (like in this case) you can simple extend it by using component composition:

```js
function StateItem({ loading, value }) {
    return <ListItem value={loading ? <Loader /> : value} />
}
```

0xDazzer: We try to keep all our components in core very simple, lean and extenciple. Here is another example, if we have Button component and you wanna add some extra functionality by using the same approach:

```js
function LoadingButton({ loading, children, ...restProps }) {
     return <Button {...restProps}>{loading ? <Loader /> : children}</Button>
}
```

0xDazzer (approved):

PR #744, opened 2025-03-12
feat: Keplr override modal error

gbarkhatov:
In case of `Keplr` override we will throw an error in `connectWallet()`

Current result: https://i.imgur.com/gnCOVsf.png

Things I don't like - initially in the design we had a different error modal. So we either match it, or keep our `ErrorModal` usage. Right now `handleError` usage is pretty simple in each of the screen, so we might do a different / more complicated UI approach for this particular use case. @0xDazzer @jonybur wdyt?

Related PR from `wallet-connector` side: https://github.com/babylonlabs-io/wallet-connector/pull/268
Related issue: https://github.com/babylonlabs-io/wallet-connector/issues/232


0xDazzer:
I think we can solve this issue without throwing any errors (no eto ne tochno)

gbarkhatov:
@jrwbabylonlab our standard error handler is different. So we might not throw an error as @0xDazzer suggests above

Question then - should this modal UI be a part of `simple-staking` or `wallet-connector`?

gbarkhatov:
@0xDazzer if this is not needed - we can close it

Issue #693, opened 2025-02-18
Potential path traversal in mempool request


jrwbabylonlab:
The fetchApi function in mempool_api.ts. The function seems to have no filter for the dot(.) . It does not prevent path traversal (Requesting an arbitrary API to fake/mess up the web service), which could cause unintended behavior, but we couldn’t find any edge case to abuse this now. We believe this is not a critical thing but, we recommend checking the user input for considering the future changes.

jonybur:
@jrwbabylonlab should we close this issue? This does not seem to be needed on simple-staking

PR #813, opened 2025-03-29
feat: adds URL validation

jonybur:
Added path traversal protection to API client to prevent potential security vulnerabilities by validating URL path components before making requests.

This prevents attackers from accessing unintended resources by using malicious inputs like `../config` or encoded variants like `%2e%2e%2fconfig`. Without these checks, user-provided addresses or transaction IDs containing path traversal sequences could potentially access unauthorized API endpoints.

Implementation includes test coverage to verify both the validation logic and integration with the mempool_api functions.

https://github.com/babylonlabs-io/simple-staking/issues/693

### DIFF
--- a/src/utils/fetch.ts
+++ b/src/utils/fetch.ts
@@ -10,6 +10,35 @@ type FetchOptions = {
 };
 
 /**
+ * Validates a URL path component to prevent path traversal attacks
+ * @param path - The path component to validate
+ * @returns The validated path component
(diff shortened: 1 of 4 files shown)


0xDazzer (commented):
This validation is useless for client side code. It only really makes sense on the server side. If server has this vulnerability - we're fucked, this FE validation doesn't help or save us

jonybur:
@0xDazzer is right, I will move this code to the API side

PR #818, opened 2025-03-30
Add polyfill for Array.toSorted

supertong:
Before:

https://github.com/user-attachments/assets/5e3f7015-af17-4de7-bd7e-a3dda80ab115

After:

https://github.com/user-attachments/assets/960cde69-c409-4c6a-a35a-6cb86e7547f4

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -21,6 +21,7 @@
         "@tanstack/react-query-next-experimental": "^5.28.14",
         "@uidotdev/usehooks": "^2.4.1",
         "bitcoinjs-lib": "6.1.5",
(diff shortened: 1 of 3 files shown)


jeremy-babylonlabs (approved):
very nice!

0xDazzer:
Nice! Can you check this? https://stackoverflow.com/questions/76593892/how-to-use-tosorted-method-in-typescript
I don't think we need add polyfill explicitly

supertong:
Hey @0xDazzer , I don't think that's going to work. It's simply adding the type for `toSorted` which we already have since we have already included `esnext` in the `lib`.

PR #831, opened 2025-04-02
Include xverse from tomo connect

jrwbabylonlab:


0xDazzer:
@jrwbabylonlab did you test it? Xverse doesn't work for me

0xDazzer:
@jrwbabylonlab for me it silently throws an error. I click on wallet icon and nothing happens. I have fixed that in my commit, but is it expected behaviour? Why Tomo throws an error that user need change network in his wallet

0xDazzer (approved):

PR #850, opened 2025-04-10
feat: defaults FP filter to inactive

jonybur:
Defaults the selector for the FP Filter list to Inactive

### DIFF
--- a/src/app/components/Staking/FinalityProviders/FinalityProviderFilter.tsx
+++ b/src/app/components/Staking/FinalityProviders/FinalityProviderFilter.tsx
@@ -1,6 +1,8 @@
 import { Select } from "@babylonlabs-io/core-ui";
+import { useCallback, useEffect, useRef } from "react";
 
 import { useFinalityProviderState } from "@/app/state/FinalityProviderState";
+import { FinalityProviderState as FinalityProviderStateEnum } from "@/app/types/finalityProviders";
 
 const options = [
   { value: "active", label: "Active" },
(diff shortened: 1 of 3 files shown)


src/app/components/Staking/FinalityProviders/FinalityProviderFilter.tsx
```diff
@@ -1,14 +1,45 @@
...
   { value: "active", label: "Active" },
   { value: "inactive", label: "Inactive" },
 ];
 
 export const FinalityProviderFilter = () => {
-  const { filter, handleFilter } = useFinalityProviderState();
+  const { filter, handleFilter, isFetching, finalityProviderMap } =
+    useFinalityProviderState();
+
+  const initialCheckDone = useRef(false);
```
jonybur, line 16: This is to only trigger the switch once

src/app/components/Staking/FinalityProviders/FinalityProviderFilter.tsx
```diff
@@ -1,14 +1,45 @@
...
+        handleFilter("status", "inactive");
+      }
+    }
+  }, [
+    filter.status,
+    hasActiveProviders,
+    handleFilter,
+    isFetching,
+    finalityProviderMap,
+  ]);
```
0xDazzer, line 42: This approach is too complicated. We need this logic only for today, so you can just hardcode the default value and update it tmr

jonybur: Fair enough

0xDazzer (approved):

src/app/state/FinalityProviderState.tsx
```diff
@@ -61,8 +61,13 @@ const FILTERS = {
     filter.status && !filter.search ? STATUS_FILTERS[filter.status](fp) : true,
 };
 
+const getDefaultFilterValue = (): "active" | "inactive" => {
+  const envValue = process.env.NEXT_PUBLIC_DEFAULT_FP_FILTER;
+  return envValue === "inactive" ? "inactive" : "active";
+};
```
0xDazzer:
```suggestion
const DEFAULT_FILTER_VALUE = process.env.NEXT_PUBLIC_DEFAULT_FP_FILTER ?? "active"
```

0xDazzer:
```
const DEFAULT_FILTER_VALUE: 'active' | 'inactive' = process.env.NEXT_PUBLIC_DEFAULT_FP_FILTER === "inactive" ? "inactive" : "active";
```

0xDazzer (approved):

PR #864, opened 2025-04-14
add source map for sentry

jeremy-babylonlabs:


package.json
```diff
@@ -4,7 +4,7 @@
   "private": true,
   "scripts": {
     "dev": "next dev -H 0.0.0.0",
-    "build": "next build",
+    "build": "next build && npm run sentry:sourcemaps",
```
0xDazzer, line 7: Are you sure we need this? I think that sentry plugin for next.js can deploy source maps automatically

0xDazzer: https://docs.sentry.io/platforms/javascript/guides/nextjs/sourcemaps/

0xDazzer (approved):

PR #869, opened 2025-04-15
add user session id in sentry

jeremy-babylonlabs:
closes #869

### DIFF
--- a/src/app/hooks/useSentrySessionId.ts
+++ b/src/app/hooks/useSentrySessionId.ts
@@ -1,0 +1,23 @@
+import * as Sentry from "@sentry/nextjs";
+import { useEffect, useState } from "react";
+
+/**
+ * useSentrySessionId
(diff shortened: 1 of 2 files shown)


jonybur (approved):

0xDazzer:
It's not clear to me why do we need this. Session IDs are automatically managed by most Sentry SDKs, as noted in [Sentry documentation](https://docs.sentry.io/platforms/python/configuration/sessions/).

jeremy-babylonlabs:
https://docs.sentry.io/platforms/python/enriching-events/identify-user/

This PR is to include the user rather than the event session. As we are using tunnel now, the user id isn't forwarded to sentry and thus the user count is not reflected correctly

0xDazzer (changes requested):
Hold

jeremy-babylonlabs:
lets close it for now

PR #904, opened 2025-04-25
fix: increase the default api retry interval to 3s

jrwbabylonlab:


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "simple-staking",
-  "version": "1.0.7",
+  "version": "1.0.9",
```
jeremy-babylonlabs, line 3: Is this correct?

jeremy-babylonlabs (approved):

0xDazzer (changes requested):
5 is a huge value

0xDazzer (approved):

PR #954, opened 2025-05-05
define clientError Class

### LINKED ISSUE #943: add clientError class and define error code

jeremy-babylonlabs:
closes #943 

Consolidated specific error codes into broader categories for simpler grouping and reporting, also refined comments for clarity

### DIFF
--- a/src/errors/clientError.ts
+++ b/src/errors/clientError.ts
@@ -1,16 +1,13 @@
-import { ErrorCode } from "./codes";
-
 interface ClientErrorOptions extends ErrorOptions {
   metadata?: Record<string, any>;
 }
 
 export class ClientError extends Error {
   constructor(
-    public errorCode: ErrorCode,
+    public readonly errorCode: string,
     message: string,
     options?: ClientErrorOptions,
   ) {
     super(message, options);
-    this.name = this.constructor.name;
   }
 }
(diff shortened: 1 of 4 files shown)


src/errors/clientError.ts
```diff
@@ -0,0 +1,16 @@
...
+}
+
+export class ClientError extends Error {
+  constructor(
+    public errorCode: ErrorCode,
+    message: string,
+    options?: ClientErrorOptions,
+  ) {
+    super(message, options);
+    this.name = this.constructor.name;
```
0xDazzer, line 14: I don't think you need this

src/errors/clientError.ts
```diff
@@ -0,0 +1,16 @@
+import { ErrorCode } from "./codes";
+
+interface ClientErrorOptions extends ErrorOptions {
+  metadata?: Record<string, any>;
+}
+
+export class ClientError extends Error {
+  constructor(
+    public errorCode: ErrorCode,
```
0xDazzer, line 9:
```suggestion
    public readonly errorCode: string,
```

src/errors/codes.ts
```diff
@@ -0,0 +1,29 @@
+export enum ErrorCode {
```
0xDazzer, line 1:
```suggestion
export const ErrorCode = {
```

src/errors/clientError.ts
0xDazzer: It's better if our error class doesn't depend on some external enum (not needed)

0xDazzer (dismissed):

PR #957, opened 2025-05-06
fix user count not tracked

### LINKED ISSUE #870: User count not tracked on Sentry

jeremy-babylonlabs:
This PR ensures proper unique user counting in Sentry
 
closes #870

Screenshot 2025-05-06 at 6 04 13 PM
Screenshot 2025-05-06 at 5 50 55 PM

### DIFF
--- a/src/hooks/useLogger.ts
+++ b/src/hooks/useLogger.ts
@@ -1,4 +1,4 @@
-import { addBreadcrumb, captureException, SeverityLevel } from "@sentry/nextjs";
+import { SeverityLevel, addBreadcrumb, captureException } from "@sentry/nextjs";
 
 type Context = Record<string, number | string | boolean> & {
   category?: string;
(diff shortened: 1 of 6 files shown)


src/hooks/useSessionManager.ts
0xDazzer, line 1: This logic is too complicated:

1. Why do we need use sessions? If user is unauthenticated we might use deviceId instead. We can create it and store inside localStorage.
2. We're going to start refactoring since next, so we can assume current codebase is a legacy. No need create any manager abstractions. You can use Sentry.setUser({ id }) directly in the code

src/hooks/useLogger.ts
```diff
@@ -40,5 +42,7 @@ const logger: Logger = {
 };
 
 export function useLogger(): Logger {
+  useSessionManager();
```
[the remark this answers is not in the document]
jeremy-babylonlabs: you're right

src/hooks/useSessionManager.ts
jeremy-babylonlabs: ok agree. we can do more complex handling in the refactor

PR #969, opened 2025-05-11
refactor all client error

### LINKED ISSUE #968: feat: refactor existing error to client error

jeremy-babylonlabs:
closes #968


src/errors/index.ts
```diff
@@ -3,11 +3,16 @@ interface ClientErrorOptions extends ErrorOptions {
 }
 
 export class ClientError extends Error {
+  public readonly metadata?: Record<string, any>;
```
0xDazzer, line 6: You shouldn't pu metadata inside ClientError class. Use logger instead

0xDazzer:
I don't think it worth to refactor errors rn. If do want refactor them you need use logger instead of putting data inside ClientError

jeremy-babylonlabs:
I there are a few missing, let me double check and let you know when its ready

src/app/context/wallet/BTCWalletProvider.tsx
```diff
@@ -137,25 +143,25 @@ export const BTCWalletProvider = ({ children }: PropsWithChildren) => {
...
-          data: {
-            network,
-            address,
-            publicKeyNoCoord,
-            walletName: await walletProvider.getWalletProviderName(),
-          },
+        logger.info("BTC wallet connected", {
+          network,
+          userPublicKey: publicKeyNoCoord?.toString() || "",
+          btcAddress: address || "",
```
jonybur, line 149: Is it possible htat address and/or publickey is undefined? If so shouldn't this be an error? (We are logging it as BTC wallet connected) (this is replicated across other error refactors as well)

0xDazzer (approved):

jonybur (approved):

PR #1062, opened 2025-06-03
fix:feature flags CI config

jeremy-babylonlabs:
we need to add it in `next.config.mjs` because we are using the FF in client side code. otherwise we would need to add prefix `NEXT_PUBLIC` which I think is not preferred


jeremy-babylonlabs:
because we will be migrating away from nextjs, i dont prefer the FF name to be dependent on the nextjs convention

jeremy-babylonlabs:
change

PR #1079, opened 2025-06-04
feat: add event bus

0xDazzer:

PR #1087, opened 2025-06-04
feat: Unisat contract UI

### LINKED ISSUE #929: Show Signing information on dApp
Unisat and other wallets is going to support verification and display of the tapscript output. 
Once that part of integration is done, we should start look into how to display the relevant information on our dApp side so that users can do a side by side comparison

This work require inputs from @DanBabylon on the design

gbarkhatov:
This PR relies on the #1079 for the cleaner data setting as
- staking after the EOI will cause the `stake event` to set the options for the `staking state`
- while closing this modal and clicking the `stake` button in `Delegations` component will cause setting the `delegation v2 state`

---

TODOs

- data flow on a proper emitter
- details to receive headers and array of key / value pairs instead of `SignPsbtOptions | something`

---

Closes #929


gbarkhatov:
Closed in favor of https://github.com/babylonlabs-io/simple-staking/pull/1153

PR #1088, opened 2025-06-05
fix: error message map

jeremy-babylonlabs:


src/app/components/Multistaking/MultistakingForm/PreviewButton.tsx
0xDazzer, line 1: You don't really need this. Just create a new validation schema for multistaking form (with new copy)

0xDazzer (changes requested):

PR #1155, opened 2025-06-20
feat: add BSNs to delegationList

jeremy-babylonlabs:


jeremy-babylonlabs:
closing because the design is changed.

PR #1165, opened 2025-06-26
fix: phase 2

jonybur:


src/ui/components/Staking/StakingModal/index.tsx
```diff
@@ -36,7 +36,7 @@ export function StakingModal() {
     reset: resetState,
     stakingStepOptions,
   } = useStakingState();
-  const { getRegisteredFinalityProvider } = useFinalityProviderBsnState();
+  const { getRegisteredFinalityProvider } = useFinalityProviderState();
```
jonybur, line 39: *

src/ui/hooks/services/useStakingService.ts
```diff
@@ -38,43 +37,24 @@ export function useStakingService() {
...
-      stakingAmountSat: btcToSatoshi(amount),
-      stakingTimelock: term,
-      feeRate: feeRate,
-    };
-    // Calculate the staking fee
-    const feeAmount = estimateStakingFee(eoiInput, feeRate);
-    return feeAmount;
-  };
+    }: Omit<FormFields, "feeAmount">) => {
+      const eoiInput = {
```
jonybur, line 48: Simplified the fee calculation given that we can have a single source of truth for the fps list

src/ui/components/Staking/StakingFormWrapper.tsx
```diff
@@ -0,0 +1,14 @@
...
+
+export function StakingFormWrapper() {
+  return (
+    <StakingState>
+      <FinalityProviderState>
+        <StakingForm />
+      </FinalityProviderState>
+    </StakingState>
+  );
+}
```
jonybur, line 14: We had lost the staking state wrapper

PR #1199, opened 2025-07-02
feat: add GA

0xDazzer:


.github/workflows/publish.yaml
```diff
@@ -55,6 +55,7 @@ jobs:
           NEXT_PUBLIC_DISABLED_WALLETS: ${{ vars.NEXT_PUBLIC_DISABLED_WALLETS }}
           S3_PUBLIC_BUCKET: ${{ vars.S3_PUBLIC_BUCKET }}
           SENTRY_ORG: ${{ vars.SENTRY_ORG }}
+          GA4_MEASUREMENT_ID: ${{ vars.GA4_MEASUREMENT_ID }}
```
filippos47, line 58: Will we have any issue for envs where this is empty (e.g., devnets)?

filippos47 (approved):

.github/workflows/publish.yaml
filippos47: Read code below, we should not.

filippos47 (approved):

PR #1217, opened 2025-07-08
feat: improve text for action

jonybur:


src/ui/common/components/Multistaking/BsnFinalityProviderField/BsnFinalityProviderField.tsx
```diff
@@ -87,14 +87,20 @@ export function BsnFinalityProviderField({ max }: Props) {
     }
   };
 
+  const actionText = useMemo(
+    () =>
+      allowsMultipleBsns
+        ? "Add BSN and Finality Provider"
+        : "Add Finality Provider",
+    [allowsMultipleBsns],
+  );
```
0xDazzer, line 96: @jonybur you don't need `useMemo` here. The expression is too simple, so `useMemo` adds  unnecessary overhead

0xDazzer (changes requested):

0xDazzer (approved):

PR #1262, opened 2025-07-18
fet: implements core-ui components for staking form

jonybur:
Replaces MultistakingForm implementation (on phase-3) with core-ui components/widgets


jeremy-babylonlabs (approved):
lgtm!

PR #1263, opened 2025-07-18
chore: clean up NEXT_PUBLIC_FF_MULTISTAKING FF

jrwbabylonlab:
This PR clean up the Feature flag `NEXT_PUBLIC_FF_MULTISTAKING` and the code that falls under it when value is false.

A new PR will be created with a new FF_NEW_UI to toggle between the phase2.5 and new phase-3 UI but with single FP support for staking


src/ui/legacy/components/Loading/Loading.tsx
```diff
@@ -1,24 +1,8 @@
-import { twJoin } from "tailwind-merge";
-
 interface LoadingProps {
   text?: string;
   noBorder?: boolean;
 }
 
-export const LoadingView: React.FC<LoadingProps> = ({ text, noBorder }) => {
```
gbarkhatov, line 8: Why do we change it in FF PR?

gbarkhatov (approved):

jeremy-babylonlabs (approved):

PR #1320, opened 2025-07-30
feat: add showExpansion condition

jeremy-babylonlabs:
This pull request updates how the UI determines whether to show the staking expansion section for delegations. The logic now checks a `canExpand` property from the backend instead of inferring expansion eligibility solely from the delegation state.

The `showExpansionSection` condition in `ActivityList.tsx` now uses the `canExpand` property from the delegation object, rather than checking if the delegation is active. This ensures that expansion is only shown when explicitly allowed by backend logic.


src/ui/common/api/getDelegationsV2.ts
```diff
@@ -116,7 +116,6 @@ const apiToDelegationV2 = (apiDelegation: DelegationV2API): DelegationV2 => {
     bbnInceptionHeight: apiDelegation.delegation_staking.bbn_inception_height,
     bbnInceptionTime: apiDelegation.delegation_staking.bbn_inception_time,
     state,
-    canExpand: apiDelegation.can_expand,
```
gbarkhatov, line 119: Is it not the case anymore?

src/ui/common/components/Activity/components/ActivityList.tsx
```diff
@@ -255,9 +255,13 @@ const transformToActivityCard = (
...
 
   // Check if expansion section should be shown
+  // 1. Feature flag enabled
+  // 2. Delegation is active
+  // 3. Delegation can expand
   const showExpansionSection =
     FeatureFlagService.IsStakingExpansionEnabled &&
-    delegation.state === DelegationV2StakingState.ACTIVE;
+    delegation.state === DelegationV2StakingState.ACTIVE &&
+    delegation.canExpand;
```
gbarkhatov, line 264: But then we rely on this

src/ui/common/context/rpc/BbnRpcProvider.tsx
```diff
@@ -1,5 +1,5 @@
 import { QueryClient } from "@cosmjs/stargate";
-import { Tendermint34Client } from "@cosmjs/tendermint-rpc";
+import { CometClient, connectComet } from "@cosmjs/tendermint-rpc";
```
gbarkhatov, line 2: Why do we switch to Comet client?

src/ui/common/components/Activity/components/ActivityList.tsx
```diff
@@ -255,9 +255,10 @@ const transformToActivityCard = (
   );
 
   // Check if expansion section should be shown
+  // 1. Feature flag enabled
+  // 2. Delegation can expand (is active logic is handled in the backend)
   const showExpansionSection =
```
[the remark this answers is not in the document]
jeremy-babylonlabs: hmm I see, ok let me add it back

PR #1342, opened 2025-08-04
feat: frozen table column

jeremy-babylonlabs:


gbarkhatov:
@jeremy-babylonlabs it has lint e2e issues

PR #1368, opened 2025-08-06
feat: implement BABY activity tab

jonybur:
https://github.com/user-attachments/assets/a7c2d434-ac2e-4ec5-a70e-55817e3aaad2


src/ui/baby/components/ActivityList/index.tsx
```diff
@@ -0,0 +1,138 @@
...
+    if (!unbondingModal.delegation) return;
+
+    try {
+      await unbond({
+        validatorAddress: unbondingModal.delegation.validator.address,
+        amount,
+      });
+      closeUnbondingModal();
+    } catch {
+      // Error is already handled in the delegation state
```
gbarkhatov, line 51: Do we need a catch then?

src/ui/baby/components/ActivityList/index.tsx
```diff
@@ -0,0 +1,138 @@
...
+              },
+          details: [],
+          optionalDetails: [
+            {
+              label: "Validator",
+              value: delegation.validator.name || delegation.validator.address,
+            },
+            {
+              label: "Commission",
+              value: `${(delegation.validator.commission * 100).toFixed(2)}%`,
```
gbarkhatov, line 80: I suggest we export this to a function (maybe in a next PR, good for now). But then let's create an issue for that so we don't forget

src/ui/baby/hooks/services/useDelegationService.ts
```diff
@@ -20,45 +21,90 @@ export interface Delegation {
...
+    });
+
+    const result = Object.values(
+      delegations.reduce(
+        (acc, item) => {
+          const validatorAddress = item.delegation.validatorAddress;
+          const delegation = acc[validatorAddress];
+
+          const isUnbonding = unbondingValidatorAddresses.has(validatorAddress);
+          const status: "active" | "unbonding" | "unbonded" = isUnbonding
```
gbarkhatov, line 74: nit: Should we extract `"active" | "unbonding" | "unbonded"` as a type?

src/ui/common/utils/wallet/bbnRegistry.ts
```diff
@@ -40,6 +44,10 @@ const protosToRegister: ProtoToRegister<any>[] = [
     typeUrl: BBN_REGISTRY_TYPE_URLS.MsgWrappedDelegation,
     messageType: epochingtx.MsgWrappedDelegate,
   },
+  {
+    typeUrl: BBN_REGISTRY_TYPE_URLS.MsgWrappedUndelegation,
+    messageType: epochingtx.MsgWrappedUndelegate,
+  },
```
gbarkhatov, line 50: Why not put
```
BBN_REGISTRY_TYPE_URLS.MsgWithdrawDelegatorReward,
    MsgWithdrawDelegatorReward
```
here?

src/ui/common/utils/wallet/bbnRegistry.ts
```diff
@@ -63,5 +71,11 @@ export const createBbnRegistry = (): Registry => {
     registry.register(proto.typeUrl, generatedType);
   });
 
+  // Register Cosmos SDK message types
+  registry.register(
```
gbarkhatov, line 75: Should be moved above
If not - should have a comment why we need it here and not as a part of protos to register

gbarkhatov (approved):

PR #1454, opened 2025-08-25
Copy update 2

jrwbabylonlab:
image

Instead of `Babylon Stats`


gbarkhatov (approved):
There are pending package.json stuff

gbarkhatov (approved):

PR #1475, opened 2025-08-26
fix: general ui issues

necipsagiro:
- fix broken back to home button and lacking footer logos in error page
  before:
  image
  after:
  image

### DIFF
--- a/src/ui/common/components/Error/GenericError.tsx
+++ b/src/ui/common/components/Error/GenericError.tsx
@@ -15,13 +15,13 @@ interface Props {
 export default function GenericError({
   title = "Oops! Something Went Wrong",
   message = `It looks like we’re experiencing a temporary issue on our end.
-  Our team is already on it, and we’ll have things back to normal as soon as possible. 
+  Our team is already on it, and we’ll have things back to normal as soon as possible.
   Please check back shortly, and thank you for your patience!`,
   image = BitcoinBlock,
 }: Props) {
   return (
     <div className="h-full min-h-svh w-full flex flex-col justify-between">
-      <SimplifiedHeader isMinimal />
+      <SimplifiedHeader />
 
       <Container className="py-12 px-6">
         <div className="flex flex-col items-center justify-center gap-8">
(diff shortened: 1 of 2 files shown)


src/ui/common/components/Logo/MobileLogo.tsx
jeremy-babylonlabs, line 1: should we put this logo to core-ui too? @jonybur

src/ui/common/components/Logo/MobileLogo.tsx
```diff
@@ -0,0 +1,20 @@
+export const MobileLogo: React.FC = () => (
+  <a href="/">
+    <svg
```
gbarkhatov, line 3: Can we use a file instead of inline `svg`?

src/ui/common/components/Error/GenericError.tsx
```diff
@@ -41,7 +41,7 @@ export default function GenericError({
           <Button
             variant="outlined"
             color="primary"
-            onClick={() => history.go()}
+            onClick={() => (window.location.href = "/")}
```
gbarkhatov, line 44: Let's not touch the `window.location.href` directly and use `router` approach

gbarkhatov (dismissed):
Approved with minor comments
@necipsagiro we can split such PRs to multiple smaller ones next time, so it's easier to review - X for menu, Y for colors, Z for navigation, etc.

src/ui/common/components/Logo/MobileLogo.tsx
necipsagiro: Logo.tsx and SmallLogo.tsx are components too, I tried to keep it consistent. Should I update them too? @gbarkhatov

jonybur (approved):

src/ui/common/components/Logo/MobileLogo.tsx
jonybur: Yeah this is a good candidate. We should have an 'icons' category in core-ui

PR #1512, opened 2025-08-29
fix(BABY): Should only claim BABY token for baby staking reward

jrwbabylonlab:
image


gbarkhatov (approved):

gbarkhatov:
@jrwbabylonlab lint error

PR #1515, opened 2025-08-29
fix: use babylon-proto-ts for btc staking

necipsagiro:
WIP

I'll make `babyTipHeightQuery` use `babylon-proto-ts` too after https://github.com/babylonlabs-io/babylon-toolkit/pull/124 merge.

### DIFF
--- a/src/ui/common/hooks/client/rpc/queries/useBbnQuery.ts
+++ b/src/ui/common/hooks/client/rpc/queries/useBbnQuery.ts
@@ -1,13 +1,6 @@
-import {
-  btclightclientquery,
-  incentivequery,
-} from "@babylonlabs-io/babylon-proto-ts";
-import {
-  QueryClient,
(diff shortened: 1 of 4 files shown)


src/ui/common/hooks/client/rpc/queries/useBbnQuery.ts
```diff
@@ -148,56 +83,22 @@ export const useBbnQuery = () => {
...
-          "Error getting Babylon chain height",
-          { cause: error as Error },
-        );
-      }
+      const { baby } = await createRPCClient({ url: BBN_RPC_URL });
+      const height = await baby.getBlockHeight();
+      return height;
     },
-    enabled: Boolean(tmClient && connected),
+    enabled: connected,
```
necipsagiro, line 93: Should we add `!isGeoBlocked && !isHealthcheckLoading` control here too?

Issue #1488, opened 2025-08-27
[UI] Show tooltip for pending stake


jrwbabylonlab:
In the BABY staking, we should show below tooltip for the two pending fields

<img width="711" height="305" alt="Image" src="https://github.com/user-attachments/assets/1e4652e0-8ba3-4be6-b463-36874af00bd8" />

1. Pending stake: Your stake will be activated in the next epoch, which takes around 1 hour
2. Pending unbonding: It will take 50 hours for the amount to be liquid

PR #1520, opened 2025-09-01
feat: add tooltip for pending states in baby staking #1488

necipsagiro:
image
image


necipsagiro:
hey @jeremy-babylonlabs,
image
seems like the tooltip box is shifted for the Hint in core-ui, should i fix it in a separate pr?

Issue #1532, opened 2025-09-03
02 - Expansion - update `ActivityList` filtering logic


gbarkhatov:
Modify the `ActivityList` component to use the new expansion visibility service for proper filtering of delegations. This will ensure the correct delegations appear in the Activity tab.

- [x] Update `src/ui/common/components/Activity/components/ActivityList.tsx`
- [x] Replace current filtering logic with calls to `useExpansionVisibilityService`
- [x] Filter out non-broadcasted VERIFIED expansions (they go to existing modal)
- [x] Include **broadcasted** VERIFIED expansions (from localStorage)
- [x] Hide original staking delegations that have broadcasted expansions
- [x] Maintain existing performance optimizations (useMemo, etc.)
- [x] Ensure backward compatibility with existing functionality
- [x] No breaking changes to existing Activity card display

- Current filter on line 55-59 filters out `EXPANDED` delegations
- Need to add additional filtering for:
  - VERIFIED expansions that aren't broadcasted
  - Original transactions that have broadcasted expansions
- Use the visibility service methods for clean, testable logic
- Maintain existing transformation and action button logic

**Dependencies**: #1530

PR #1543, opened 2025-09-04
feat: expansion activity

gbarkhatov:
- solves the activity tab problem for exansion
- split the logic in activity list
- formats the `src/ui/common/components/Stats/Stats.tsx` because of `npm run format:fix`

Closes https://github.com/babylonlabs-io/simple-staking/issues/1532

### DIFF
--- a/src/ui/common/hooks/services/useActivityDelegations.ts
+++ b/src/ui/common/hooks/services/useActivityDelegations.ts
@@ -1,0 +1,75 @@
+import { useBTCWallet } from "@/ui/common/context/wallet/BTCWalletProvider";
+import { useDelegationService } from "@/ui/common/hooks/services/useDelegationService";
+import { useExpansionVisibilityService } from "@/ui/common/hooks/services/useExpansionVisibilityService";
+import { useStakingManagerService } from "@/ui/common/hooks/services/useStakingManagerService";
+import { useFinalityProviderState } from "@/ui/common/state/FinalityProviderState";
+
+import {
+  ActivityCardData,
+  useActivityCardTransformation,
+} from "./useActivityCardTransformation";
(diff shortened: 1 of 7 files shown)


src/ui/common/hooks/services/useActivityDelegations.ts
```diff
@@ -0,0 +1,75 @@
...
+    publicKeyNoCoord,
+  );
+
+  // Step 2: Apply expansion visibility filtering
+  const visibleDelegations = getActivityTabDelegations(validatedDelegations);
+
+  // Step 3: Transform to activity card data
+  const activityData: ActivityCardData[] = useActivityCardTransformation(
+    visibleDelegations,
+    finalityProviderMap,
```
jonybur, line 53: If finalityProviderMap mutates, the memo in useActivityCardTransformation won't update

src/ui/common/hooks/services/useActivityDelegations.ts
```diff
@@ -0,0 +1,75 @@
...
+    openConfirmationModal,
+    isStakingManagerReady,
+  );
+
+  return {
```
jonybur, line 58: Perhaps worth memoing this return object, otherwise this might trigger rerenders

jonybur (commented):
Left some comments

src/ui/common/hooks/services/useActivityDelegations.ts
gbarkhatov: Done

src/ui/common/hooks/services/useActivityDelegations.ts
gbarkhatov: It is already there in the deps

